PR #1, opened 2016-11-29
Initialize project

jaywink:
A basic project skeleton with some basic Django settings etc.

Questions:
* [x] Should we place the `requirements*.*` files in a `requirements/` folder?
* [x] Should the Django app actually be called `parkinghub` not `parkkihub`? If the Finnish language version is fine, better keep it aligned with the repo name.

TODO:
* [x] Needs LICENSE file!


tuomas777:
About requirements folder, I prefer the current layout without a folder, I think it's more common and easier to use and 4 files isn't that much yet.

tuomas777:
About naming the app, urgh, these things really suck :) I think the name should be in English. Some Helsinki APIs seem to follow a convention where the app's name is it's main model's name pluralized, so in this case I guess it would be something like parkings. But yeah, I find that convention a bit odd so dunno :D

README.md
```diff
@@ -1,3 +1,65 @@
...
+
+Use `pip-tools` to update the `requirements*.txt` files.
+
+    pip install -U pip  # pip-tools needs pip==6.1 or higher (!)
+    pip install pip-tools
+    
+When you change requirements, set them in `requirements.in` or `requirements-dev.in`. Then run:
+
+    pip-compile --output-file requirements.txt requirements.in
+    pip-compile --output-file requirements-dev.txt requirements-dev.in
```
jaywink: -o, --output-file TEXT      Output file name. Required if more than one
                                input file is given. Will be derived from input
                                file otherwise.

Good point 👍

setup.py
```diff
@@ -0,0 +1,11 @@
...
+
+setup(
+    name='parkkihubi',
+    version='0.0.0',
+    license='???',
+    packages=find_packages(),
+    include_package_data=True,
+    install_requires=[l for l in open("requirements.txt", "rt").readlines() if l and not l.startswith("#")],
+    zip_safe=False,
+)
```
jaywink: I'll remove it - we probably wont need to build wheels.

jaywink:
As discussed IRL, `parkkihubi` is probably good to keep as the repo is named that. However, actual features should live in logically named Django applications.

jaywink:
Final approval before merge? :)

PR #2, opened 2016-12-02
Add parkings app with some models

jaywink:
Parking, Address and Operator models will form base of data storage requirements.

Currently no logic tied to these, that will come with REST framework addition.


parkings/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .addresses import *
+from .operators import *
+from .parkings import *
```
jaywink: Odd, I've seen the complete opposite ;) For example if the `parking.py` would contain in theory many different parking models, would you still use singular?

But changing for this 👍

juyrjola: Yeah, we seem to have standardized on having the app name in plural and these kinds of break-out files in singular.

Also, according to the Zen of Python, _explicit is better than implicit_, so import the classes you need explicitly instead of `*`. This saves some time when someone tries to figure out where the classes are actually defined.

jaywink: Sure, can do 👍 It's quite common imho to use the combo of `*` import in `__init__.py` and then define `__all__` in the files themselves (to not create that pollution), but explicit without __all__ is probably clearer as said.

parkings/tests/test_models.py
```diff
@@ -0,0 +1,31 @@
+import datetime
+
+from django.contrib.gis.geos import Point
+from django.utils.timezone import now
+
+from parkings.models import Address, Operator, Parking
+
+
+class TestAddressModel(object):
```
jaywink: Sure, can do 👍 

Pytest actually does recommend [grouping tests in classes](http://doc.pytest.org/en/latest/getting-started.html#grouping-multiple-tests-in-a-class) for maintainability. It helps especially when the project becomes more complex and the amount of tests is higher. But functions are ok too, changed 👍

jaywink:
@juyrjola do HKI projects have a preference for single quotes vs double quotes? Still time to enforce something like that if there is a guideline.

juyrjola:
@jaywink We don't have a written preference yet, but I've used double quotes for strings that are visible to the user (such as full English sentences) and single quotes for identifiers and such.

PR #3, opened 2016-12-09
Initial APIs

tuomas777:
Added initial admin UI and initial API implementations (write-only operator and read-only internal).

### DIFF
--- a/parkings/models/parking.py
+++ b/parkings/models/parking.py
@@ -1,5 +1,6 @@
 from django.contrib.gis.db import models
 from django.core.validators import RegexValidator, MinValueValidator, MaxValueValidator
+from django.utils.timezone import localtime
 from django.utils.translation import ugettext_lazy as _
 
 from parkings.models.address import Address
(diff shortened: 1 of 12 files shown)


parkings/models/parking.py
```diff
@@ -37,3 +38,8 @@ class Parking(TimestampedModelMixin, UUIDPrimaryKeyMixin):
     zone = models.IntegerField(verbose_name=_("zone number"), validators=[
         MinValueValidator(1), MaxValueValidator(3),
     ])
+
+    def __str__(self):
+        start = localtime(self.time_start).replace(tzinfo=None)
+        end = localtime(self.time_end).time().replace(tzinfo=None)
+        return "%s -> %s" % (start, end)
```
jaywink: I guess to make this useful a registration number might be nice here.

parkings/api/internal/__init__.py
```diff
@@ -0,0 +1 @@
+from .parking import InternalAPIParkingViewSet
```
tuomas777: Yeah in the current implementation there isn't, removing.

parkings/models/parking.py
tuomas777: True, adding.

PR #11, opened 2016-12-29
Allow time_end to be PATCHed after time editable is over

tuomas777:


parkings/api/operator/parking.py
```diff
@@ -57,6 +57,10 @@ def update(self, instance, validated_data):
         return instance
 
     def validate(self, data):
+        if self.instance and (now() - self.instance.created_at) > settings.PARKINGS_TIME_EDITABLE:
+            if set(data.keys()) != {'time_end'}:
+                raise exceptions.PermissionDenied(_('Other than PATCHing "time_end" not allowed anymore.'))
```
jaywink: Is `PermissionDenied` correct to return here? Somehow I think `VaidationError` would make more sense and it would result in the correct 400 status code?

Also, for error string, maybe: `Grace period has passed. Only "time_end" can be updated via PATCH.` ?

jaywink: Talked IRL that 403 was seen in previous discussions in similar API situations as the correct code to return in this kind of validation situation where "train has passed" even though this has absolute nothing to do with permissions.

Still, error message needs some language correctness 👍 Once the error message has been tweaked, ready to 🚢

PR #15, opened 2017-02-03
Operator API changes and docs update

tuomas777:


docs/api/operator.yaml
```diff
@@ -162,50 +218,106 @@ paths:
...
+                type: integer
+                enum: [1, 2, 3]
       responses:
         200:
-          description: The updated parking
+          description: The parking was updated successfully
           schema:
             $ref: '#/definitions/Parking'
         401:
+          description: Unauthorized
```
mingfeng: Why not `$ref: '#/responses/Unauthorized'`?

parkings/api/operator/parking.py
```diff
@@ -1,53 +1,22 @@
...
-from parkings.models import Address, Operator, Parking
+from parkings.models import Operator, Parking
 
-from ..common import AddressSerializer, ParkingFilter
+from ..common import ParkingFilter
 
 
 class OperatorAPIParkingSerializer(serializers.ModelSerializer):
     status = serializers.ReadOnlyField(source='get_state')
-    address = AddressSerializer(allow_null=True, required=False)
```
mingfeng: So parkings have addresses but we don't expose it to the REST API?

mingfeng:
Looks solid to me :+1:

parkings/api/operator/parking.py
tuomas777: Yeah, well, we will probably remove addresses completely, but that is not sure yet, so I started with removing it from Operator API, which is the one that actually matters ATM. :)

PR #16, opened 2017-03-28
Parking areas

maxg0:
This contains the new parking areas model and an importer that uses the kartta.hel.fi WFS as a source, as well as a sanity test for the parser (but not for checking if the WFS still works).


parkings/importers/parking_areas.py
```diff
@@ -0,0 +1,201 @@
...
+                progress = 100 * ((index + 1) / area_dicts_count)
+                sys.stdout.write('\rCreating areas %d%%' % progress)
+
+            try:
+                parking_area = ParkingArea.objects.get(
+                    external_id=area_dict['external_id'],
+                )
+                self._update_parking_area(area_dict, parking_area)
+            except ParkingArea.DoesNotExist:
+                parking_area = self._create_parking_area(area_dict)
```
tuomas777: I think using `get_or_create()` would be cleaner, something like:

```
created, parking_area = ParkingArea.objects.get_or_create(
   external_id=area_dict['external_id'],
   defaults={ ... }
)
if not created:
    self._update_parking_area( ... )
```

And then you need `save()` inside the update method, which actually makes more sense as you might skip saving if there is no overwrite.

parkings/importers/parking_areas.py
```diff
@@ -0,0 +1,201 @@
...
+            areas=area_dict['areas'],
+            space_amount_estimate=area_dict['space_amount_estimate'],
+        )
+        self.created += 1
+        return parking_area
+
+    def _update_parking_area(self, area_dict, parking_area):
+        space_amount_estimate = area_dict['space_amount_estimate']
+        if space_amount_estimate:
+            parking_area.space_amount_estimate = space_amount_estimate
```
tuomas777: Hmm, this will update the estimate regardless of `self.overwrite` and that is probably not intentional?

parkings/models/parking_area.py
```diff
@@ -0,0 +1,38 @@
...
+    areas = models.MultiPolygonField(srid=3879)
+
+    '''
+    This is a rough capacity estimate of how many cars might fit into the
+    parking area.
+    '''
+    space_amount_estimate = models.PositiveSmallIntegerField(
+        null=True,
+        blank=True,
+    )
```
tuomas777: `verbose_name` is missing from this and `areas` field above.

parkings/models/parking_area.py
```diff
@@ -0,0 +1,38 @@
...
+    This is for whatever ID the external system that this parking lot was
+    imported from has assigned to this lot. There is no guarantee that it will
+    always be a number or that it will be unique, especially if multiple
+    sources are to be used.
+    '''
+    external_id = models.CharField(
+        max_length=32,
+        verbose_name=_('external ID'),
+        db_index=True,
+    )
```
maxg0: True, things should fail when we're trying to add the bad data, not when we accidentally get two results for the same query when we're expecting only one.

parkings/importers/parking_areas.py
maxg0: Nice catch! Actually a few lines down this is done correctly so this is just leftover code.

requirements-dev.txt
```diff
@@ -4,6 +4,7 @@
 #
 #    pip-compile --output-file requirements-dev.txt requirements-dev.in
 #
+appdirs==1.4.3            # via setuptools
```
maxg0: I removed the offending lines, reinstalled the whole virtualenv and it still works.

tuomas777:
SHIP IT! :shipit:

PR #18, opened 2017-04-24
Public API and Parking Areas

maxg0:
Adds a public API and in it, parking areas, in a somewhat GeoJSON compatible format.


parkings/api/public/parking_area.py
```diff
@@ -0,0 +1,26 @@
...
+    class Meta:
+        model = ParkingArea
+        geo_field = 'wgs84_areas'
+        fields = (
+            'id',
+            'space_amount_estimate',
+        )
+
+
+class PublicAPIParkingAreaViewSet(viewsets.ModelViewSet):
```
tuomas777: This should inherit from `viewsets.ReadOnlyModelViewset`

tuomas777:
LGTM!

PR #19, opened 2017-04-25
Api error codes

tuomas777:
Added error code to 403 API errors to make them machine-readable.

The code is returned in `code` field in 403 response body alongside
the default `detail` field.

current codes:

`unknown_error`: Fallback value, should not be used normally
`permission_denied`: for DRF PermissionDenied()
`grace_period_over`: Parking edit grace period has passed


parkings/tests/api/internal/test_parking.py
```diff
@@ -49,7 +49,7 @@ def check_parking_data_matches_parking_object(parking_data, parking_obj):
 def test_other_than_staff_cannot_do_anything(unauthenticated_api_client, operator_api_client, parking):
     urls = (list_url, get_detail_url(parking))
     check_method_status_codes(unauthenticated_api_client, urls, ALL_METHODS, 401)
-    check_method_status_codes(operator_api_client, urls, ALL_METHODS, 403)
+    check_method_status_codes(operator_api_client, urls, ALL_METHODS, 403, 'permission_denied')
```
maxg0: This should be a kwarg.

PR #20, opened 2017-04-26
Parking area statistics public API

maxg0:
Adds a real-time public API that shows the amount of parkings per parking area.

### DIFF
--- a/parkings/api/public/parking_area_statistics.py
+++ b/parkings/api/public/parking_area_statistics.py
@@ -3,6 +3,8 @@ from rest_framework import serializers, viewsets
 
 from parkings.models import Parking, ParkingArea
 
+from ..common import WGS84InBBoxFilter
+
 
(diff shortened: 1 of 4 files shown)


maxg0:
Actually I'm merging this into the parking_area endpoint, will make another PR later.

parkings/api/public/parking_area_statistics.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ParkingAreaStatisticsSerializer(serializers.ModelSerializer):
+    current_parking_count = serializers.SerializerMethodField()
+
+    def get_current_parking_count(self, area):
+        count = Parking.objects.filter(
+            parking_area=area,
+            time_end__gte=timezone.now(),
+            time_start__lte=timezone.now(),
+        ).count()
```
tuomas777: This results in gazillion db queries

PR #21, opened 2017-04-28
Remove obsolete stuff

tuomas777:
* Removed `Address` model and `device_identifier`, `resident_code` and `special_code` from `Parking`

* Disabled GET method operations from operator API

* Fixed ParkingFactory location field


maxg0:
LGTM

PR #22, opened 2017-04-28
Public api enhancements

tuomas777:


parkings/api/public/urls.py
```diff
@@ -6,7 +6,7 @@
 
 router = DefaultRouter()
 router.register(r'parking_area', PublicAPIParkingAreaViewSet)
```
maxg0: Should this have a `base_name` as well for consistency?

maxg0: Other than that, LGTM, can be merged.

PR #24, opened 2017-05-08
Fix some verbose names

tuomas777:


parkings/migrations/0008_fix_some_verbose_names.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('parkings', '0007_rename_parking_area_fields'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='parkingarea',
+            name='capacity_estimate',
+            field=models.PositiveSmallIntegerField(blank=True, null=True, verbose_name='capasity estimate'),
```
maxg0: Except this. Should be "capacity estimate".

PR #28, opened 2017-06-08
Public API parking endpoint

tuomas777:
Added parking endpoint to public API.

Parkings in the public side contain fields `id`, `parking_area`, `location`,
`time_start`, `time_end` and `zone`.

There is a setting `PARKKIHUBI_TIME_PARKINGS_HIDDEN` which is used to
determine how long new parkings are excluded from the public API
(default 7 days).

Available filters:
  - `time_start_lte`
  - `time_start_gte`
  - `time_end_lte`
  - `time_end_gte`
  - `parking_area` (id, allows multiple comma-separated values)
  - `zone` (allows multiple comma-separated values)
  - `in_bbox` (four comma-separated values)


maxg0:
LGTM :shipit:

PR #31, opened 2017-08-23
Disable registration number validation

tuomas777:
Disable registration number validation because it was too restrictive, getting it right for all possibilities would probably be hideous, and it doesn't do that much good anyway.

The PR contains also unrelated minor string presentation tuning.


maxg0:
LGTM :shipit:

PR #33, opened 2017-08-29
Optional parking end time

tuomas777:
Change parking end time optional as it turned out it is not always available when creating a parking.

While at it, also changed time filters to use the double underscore notation from Django (`time_end_lte` -> `time_end__lte`). Reasons being that this is the default behaviour of django-filter package, and also using at least somewhat common conventions is always better than inventing your own IMO.


parkings/models/parking.py
```diff
@@ -46,7 +49,14 @@ def __str__(self):
         return "%s -> %s (%s)" % (start, end, self.registration_number)
 
     def get_state(self):
-        if self.time_start <= now() <= self.time_end:
+        current_timestamp = now()
+
+        if self.time_end is None:
```
mingfeng: The code is a bit too nested  IMO, maybe:
```
if self.time_start > current_timestamp:
    return Parking.NOT_VALID

if self.time_end and self.time_end < current_timestamp:
    return Parking.NOT_VALID

return Parking.VALID

```

parkings/tests/test_models.py
```diff
@@ -19,3 +21,28 @@ def test_parking_instance_creation():
...
+
+@pytest.mark.django_db
+@override_settings(TIME_ZONE='Europe/Helsinki')
+def test_parking_str(parking_factory):
+    parking = parking_factory(
+        time_start=datetime.datetime(2014, 1, 1, 6, 0, 0, tzinfo=utc),
+        time_end=datetime.datetime(2016, 1, 1, 7, 0, 0, tzinfo=utc),
+        registration_number='ABC-123',
+    )
+    assert all(parking.__str__().count(val) == 1 for val in ('2014', '2016', '8', '9', 'ABC-123'))
```
mingfeng: minor, but I would suggest to avoid using dunder methods when possible. Maybe `str(parking).count(val)`

PR #34, opened 2017-08-31
Remove public API parking endpoint

tuomas777:


suutari-ai:
Why is code coverage decreasing so much in parkings/factories/parking.py ? Maybe there is some code that becomes unused by this PR? If that's the case, then it should be removed too.

suutari-ai:
Is it so that the HistoryParking / history_parking_factory is no longer used?

tuomas777:
Yes that was the case, good catch! But that isn't the case anymore, as HistoryParkingFactory is used in the other PR just merged to master :)

PR #36, opened 2017-09-11
parking areas importer: Fix namespace of avoindata

suutari-ai:
It seems that the avoindata namespace has changed from http to https in
the received XML content.  Amend the importer so that it can import the
data again.


vikoivun:
Everything hel.fi has converted to https, so looks sane. Although conversion of xml namespaces on the fly is silly. Probably a geoserver thing?.

PR #39, opened 2017-09-20
Update requirements and switch to Prequ

suutari-ai:
See the commit messages for details.


tuomas777:
Is there some concrete reason for using Prequ on this project? (in addition to it being superior to pip-tools in every possible way of course ;) )

suutari-ai:
IMHO `prequ check` is very good check to do for the requirements. It's kind of an unit test for checking that requirements.txt is up-to-date. That will protect from changes like when someone edits just requirements.in, but doesn't remember to recompile the requirements or when requirements.txt is edited by hand to contain conflicting requirements (e.g. some version of a package is upgraded, but another package is not compatible with the newer version).

PR #41, opened 2017-09-21
Revert "Hide registration numbers in admin"

suutari-ai:
Hiding of the registration numbers was only a temporary need.  Make them
visible again.

This reverts commit e9d90b911f6645d44d405e9951c84f400a5f8151.


tuomas777:
LGTM!

PR #42, opened 2017-10-05
Add parking query API for PASI

suutari-ai:
* Add API endpoint for getting parkings by registration number

* Add normalized registration number to Parking object and fill it with
  a data migration.


parkings/api/enforcement/active_parkings.py
```diff
@@ -0,0 +1,66 @@
...
+            all_parkings_by_reg_num,
+            starts_before=time,
+            ends_after=time)
+        if not active_parkings:
+            active_within_limit = _filter_time_range(
+                all_parkings_by_reg_num,
+                starts_before=time,
+                ends_after=(time - datetime.timedelta(days=1)))
+            return active_within_limit[:1]
+        return active_parkings
```
tuomas777: It would be nicer if `FilterSet` from `django-filter` package was used instead of overriding `get_queryset()`. One obvious benefit would be consistent parsing of filtering values instead of needing to write your own `parse_time()`.

parkings/api/enforcement/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.conf.urls import include, url
+
+from .active_parkings import ActiveParkingsView
+
+urls = [
+    url(r'^active-parkings/(?P<registration_number>.+)/',
```
tuomas777: Convention on this project's (and on other HKI projects as well) API endpoint names is to use underscores instead of dashes and singular form instead of plural (so the name should be `active_parking`). Of course this isn't a public API so it doesn't matter that much, but I would still vote for consistency unless there are valid reasons to break it.

tuomas777: About this endpoint in general, is there a reason why it works like `/active-parkings/<reg_num>/?time=<time>` instead of `/active-parkings/?reg_num=<reg_num>&time=<time>`?

The latter would be the "default" way of doing this in DRF, and I it would be more easily extended later if/when needed. For example we could easily add a way to query multiple registration numbers on one request (`/?reg_num=<reg_num>,<reg_num>,<reg_num>` for example), which is something we very well might need in the future.

parkings/migrations/0015_fill_normalized_reg_nums.py
```diff
@@ -0,0 +1,32 @@
...
+    parkings_to_process = parking_model.objects.filter(
+        Q(normalized_reg_num=None) | Q(normalized_reg_num=''))
+    for parking in parkings_to_process:
+        parking.normalized_reg_num = Parking.normalize_reg_num(
+            parking.registration_number)
+        parking.save(update_fields=['normalized_reg_num'])
+
+
+def noop(apps, schema_editor):
+    pass
```
tuomas777: There is `RunPython.noop` for this purpose :)

suutari-ai: > The existing initial internal API was meant for this exact purpose, so it could and should be removed now then :)

OK then. I didn't want to change it, because I wasn't able to find any documentation where it's used. Bad luck, since it would have been what was needed.

suutari-ai:
> The existing initial internal API was meant for this exact purpose, so it could and should be removed now then :)

OK then. I didn't want to change it, because I wasn't able to find any documentation where it's used. Bad luck, since it would have been what was needed.

parkings/migrations/0015_fill_normalized_reg_nums.py
suutari-ai: Thanks, will use that then.

parkings/api/enforcement/urls.py
suutari-ai: The reason to put the reg_num to path instead of query parameter was that it's mandatory.

parkings/api/enforcement/active_parkings.py
```diff
@@ -0,0 +1,66 @@
...
+from django.utils import timezone
+from rest_framework import generics, permissions, serializers
+
+from ...authentication import ApiKeyAuthentication
+from ...models import Parking
+from ..utils import parse_time
+
+
+class ActiveParkingsSerializer(serializers.ModelSerializer):
+    operator_name = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
suutari-ai: Good to know, thanks!

parkings/api/enforcement/active_parkings.py
suutari-ai: I agree. I'll refactor it to use FilterSet.

suutari-ai:
OK, the requested changes are now done.

@tuomas777: Good review. I learned some good stuff from your comments. Thanks! :smile:

@jaywink Can you recheck now, please? I guess @tuomas777 is unavailable right now...

suutari-ai:
I'll create another PR for removing the internal API, if it's really unused now. (I'd like to double check that from @tuomas777, so not including it to this PR.)

parkings/api/enforcement/active_parking.py
```diff
@@ -0,0 +1,111 @@
...
+        that has the latest ending time.
+
+        :type queryset: django.db.models.QuerySet
+        :type name: str
+        :type value: datetime.datetime
+        """
+        time = value if value else timezone.now()
+        active_parkings = _filter_time_range(
+            queryset, starts_before=time, ends_after=time)
+        if not active_parkings:
```
jaywink: Personally I would prefer returning early and keeping the code flatter, ie:

    if active_parkings:
        return active_parkings
    # Do other stuff and return the result of that

parkings/api/enforcement/active_parking.py
```diff
@@ -0,0 +1,111 @@
...
+            active_within_limit = _filter_time_range(
+                queryset,
+                starts_before=time,
+                ends_after=(time - datetime.timedelta(days=1)))
+            return active_within_limit.order_by('-time_end')[:1]
+        return active_parkings
+
+
+def _filter_time_range(queryset, starts_before, ends_after):
```
jaywink: I think a customer `ParkingQuerySet` would be cleaner, then using it as a manager to `Parking`. Then you could just do:

    queryset = queryset.filter_time_range(starts_before, ends_after)

parkings/api/enforcement/active_parking.py
```diff
@@ -0,0 +1,111 @@
...
+    after the ends_after value or have no ending time set.
+
+    :type queryset: django.db.models.QuerySet
+    :type starts_before: datetime.datetime
+    :type ends_after: datetime.datetime
+    :rtype: django.db.models.QuerySet
+    """
+    return queryset.filter(
+        Q(time_start__lte=starts_before) &
+        Q(time_end__gte=ends_after) | Q(time_end__isnull=True))
```
jaywink: The `OR` would be nice to be inside parenthesis for extra clarity and to avoid future accidental problems.

parkings/api/enforcement/active_parking.py
suutari-ai: Yeah, that's nicer.. will do that.

parkings/api/enforcement/active_parking.py
suutari-ai: Good idea.

parkings/api/enforcement/active_parking.py
suutari-ai: Oops, good catch. It was meant to be so.

suutari-ai:
@jaywink Did the requested changes. Also renamed "active parkings" to "valid parkings" to avoid using two different terms for the same thing.

PR #43, opened 2017-10-06
Remove internal API and document enforcement API

suutari-ai:
Since the internal API is not used anymore (as per @tuomas777's comments
in PR #42), remove it as unnecessary.

Document the enforcement API while at it.


suutari-ai:
Closing and creating a new PR since, codecov doesn't seem to work very well for when re-pushing to same branch...

PR #45, opened 2017-12-07
enforcement/valid_parking: Tune the delay of invalid visibility

suutari-ai:
The valid_parking endpoint returns the last parking even if it is no
longer valid, when there is no valid parkings, but the latest parking is
not too old.  The limit used to be 1 day, but that was too much.  Change
that limit to 15 minutes as requested.


parkings/api/enforcement/valid_parking.py
```diff
@@ -76,8 +76,8 @@ def filter_time(self, queryset, name, value):
         valid_parkings = queryset.valid_at(time)
         if valid_parkings:
             return valid_parkings
-        day_ago = time - datetime.timedelta(days=1)
-        valid_within_limit = queryset.starts_before(time).ends_after(day_ago)
+        limit = time - datetime.timedelta(minutes=15)
```
tuomas777: Not exactly related to this change, but the time could be in settings file the same way there is `PARKKIHUBI_TIME_PARKINGS_EDITABLE`

PR #46, opened 2017-12-12
api/enforcement: Add endpoint for operators

suutari-ai:
Add API endpoint for listing the operators.

Also add operator identifier to the valid_parking endpoint so that it is
possible to link them properly (not just by name).

### DIFF
--- a/parkings/api/enforcement/urls.py
+++ b/parkings/api/enforcement/urls.py
@@ -1,9 +1,11 @@
 from django.conf.urls import include, url
 from rest_framework.routers import DefaultRouter
 
+from .operator import OperatorViewSet
 from .valid_parking import ValidParkingViewSet
 
 router = DefaultRouter()
+router.register('operator', OperatorViewSet, base_name='operator')
 router.register('valid_parking', ValidParkingViewSet,
                 base_name='valid_parking')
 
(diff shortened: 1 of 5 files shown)


parkings/api/enforcement/urls.py
```diff
@@ -1,9 +1,11 @@
 from django.conf.urls import include, url
 from rest_framework.routers import DefaultRouter
 
+from .operator import OperatorViewSet
 from .valid_parking import ValidParkingViewSet
 
 router = DefaultRouter()
+router.register('operator', OperatorViewSet, base_name='operator')
```
tuomas777: `base_name` not needed here.

parkings/tests/api/enforcement/test_valid_parking.py
```diff
@@ -75,7 +75,7 @@ def check_parking_data_keys(parking_data):
         'id', 'created_at', 'modified_at',
         'registration_number',
         'time_start', 'time_end', 'zone',
-        'operator_name',
+        'operator', 'operator_name',
```
tuomas777: Could also check the contents of this field.

parkings/api/enforcement/urls.py
suutari-ai: I like the explicit base_name more. It's kind of "magic" when its determined from the model behind the serializer's queryset.

parkings/tests/api/enforcement/test_valid_parking.py
suutari-ai: Right. Does it now.

PR #49, opened 2018-02-09
Add monitoring API and JWT auth for Dashboard

suutari-ai:
* Add new model Region, which is meant for larger areas than the current
  Parking Areas are.  The regions could be set up to cover all parking
  spot positions in Helsinki, e.g. by making regions from city parts
  such as Kamppi and Kruunuhaka.

* Add commands for importing Region data and fill regions to existing
  parkings.

* Add a new API namespace "monitoring" with couple new API endpoints:

  - Regions -- To get info about regions (name, area, capacity...)
  
  - Region statistics -- Parking statistics for regions at given time

* Add Two Factor JWT authentication implemented with Django Rest
  Framework JWT 2FA (see https://github.com/suutari/drf-jwt-2fa).

* Reconfigure API authentication classes so that its easier to
  experiment with the API endpoints while developing. Now, in
  DEBUG mode, the session auth is also accepted and then
  Django Admin login can be used to create a session.

* Add CORS headers so that API can be accessed by web browsers.

This is based on the PR #48, because the imports are formatted according
to the new isort settings in that PR.


parkings/tests/api/monitoring/test_region_statistics.py
```diff
@@ -0,0 +1,88 @@
...
+
+def test_with_many_parkings_and_specified_time(monitoring_api_client):
+    (parkings, regions) = create_parkings_and_regions(
+        parking_count=20, region_count=5)
+
+    # Pick a point in time that is in between the parking times
+    earliest_start_time = min(x.time_start for x in parkings)
+    earliest_end_time = min(x.time_end for x in parkings)
+    latest_end_time = max(x.time_end for x in parkings)
+    range_start = max(earliest_start_time, earliest_end_time)
```
tuomas777: Not sure how earliest start time could ever be after earliest end time :)

Also Faker has `date_time_between()` which could probably help on this(?)

parkkihubi/settings.py
```diff
@@ -159,7 +159,7 @@
     'ALLOWED_VERSIONS': ('v1',),
     'DEFAULT_FILTER_BACKENDS': ('django_filters.rest_framework.DjangoFilterBackend',),
     'EXCEPTION_HANDLER': 'parkings.exception_handler.parkings_exception_handler',
-    'PAGE_SIZE': 20,
+    'PAGE_SIZE': 100,
```
tuomas777: This also affects the public API, is that something we want?

suutari-ai: Yes, it will. IMHO 100 is still much better default. It can be changed with page_size param anyway.

parkings/tests/api/monitoring/test_region_statistics.py
suutari-ai: Yeah, you're right. earliest_end_time is always the maximum of the two. I'll simplify it. Will also check if date_time_between would be good here.

parkings/management/commands/fill_parking_regions.py
```diff
@@ -0,0 +1,71 @@
...
+            .exclude(location=None)
+            .filter(region=None)
+            .order_by('created_at'))
+        count = parkings.count()
+
+        if not count:
+            show_info("Nothing to do")
+            return
+
+        block_count = int(max(count / block_size_target, 1))
```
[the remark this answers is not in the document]
suutari-ai: The real parking counts are in millions, so this doesn't really matter.

suutari-ai:
@tuomas777 Thanks for the review! I've now fixed the couple typos and improved the test case to use `fake.date_time_between`.

Didn't make changes to the block size calculation of `fill_parking_regions`, nor the page size, because I think those should be OK.

suutari-ai:
Update: Fixed a minor issue with the `fill_parking_regions` command I just found. (It was missing `nargs='?'` for the `block_size_target` argument and therefore the the default value was never used, but rather a value was always required.)

PR #50, opened 2018-02-12
Add a script to fix invalid geometries

suutari-ai:
Some of the (imported) geometries might be invalid.  This causes
problems, because it makes many GIS database queries fail .  This
command can be used to fix those geometries.


parkings/management/commands/fix_geometries.py
```diff
@@ -0,0 +1,112 @@
...
+            show_debug("  after MakeValid:   {}".format(
+                _geom_info(obj.valid_geom, verbosity)))
+            fixed_geom = _to_multipolygon(obj.valid_geom)
+            show_debug("  after conversion:  {}".format(
+                _geom_info(fixed_geom, verbosity)))
+            if dry_run:
+                show_info("  Not saved, since doing dry-run.")
+            else:
+                setattr(obj, field_name, fixed_geom)
+                obj.save()
```
tuomas777: Could have `update_fields=(field_name,)` ?

parkings/tests/test_fix_geometries.py
```diff
@@ -0,0 +1,143 @@
...
+    timestamps_after = sorted(
+        ParkingArea.objects.values_list('modified_at', flat=True))
+    if dry_run:
+        assert timestamps_after == timestamps_before
+    else:
+        assert timestamps_after != timestamps_before
+
+    areas_after = get_areas(ParkingArea.objects.all())
+
+    assert areas_after == areas_before, "Areas (m^2) shouldn't change"
```
tuomas777: Are the actual area sizes floats? If they are, I don't think they should be compared like this, at least in theory :)

suutari-ai: Yes, they are floats. You're right, maybe I should convert them to Decimals or ints after the rounding.

parkings/management/commands/fix_geometries.py
suutari-ai: Then modification time wouldn't change, but IMHO it should, since the area has been changed.

tuomas777: Ok, what about `update_fields=(field_name, 'modified_at')` then ? :)

suutari-ai: What's the difference then?

suutari-ai:
Updated:
 * get_areas now returns the area sizes formatted as strings so no float comparison is needed anymore
 * added `order_by('pk')` to the command to make sure that the resulting output is always in same order and the test doesn't randomly fail
 * No changes for the save, since IMHO it's better to not mask any related field updates if the model happens to have some kind of caching bound to save method which would calculate some other fields based on the geometry fields.

PR #52, opened 2018-02-14
Dashboard: First version

suutari-ai:
Implement a dashboard JS application for monitoring parkings in
Parkkihubi.

Following features are currently implemented:

 * Login with Two Factor Authentication using JWT tokens

 * A Map showing parking regions (ParkingRegionMap)

    - Map shows usage of each region by a color from green to red

    - Map shows more info about a region by clicking a region

    - Regions can be selected by clicking them on the map

 * Region can be selected or deselected from a dropdown (RegionSelector)

 * Possibility to select a date and time of interest; by default uses
  the current time (TimeSelect)

   - Time updates automatically every 5 minutes

There is also some static non-updating dummy UI components shown to
demonstrate features that will be implemented in the future:

  * A table that will list currently active parkings

  * A bar diagram that will show histogram of the parking counts from
   last 24h (or whatever time range is selected)

Layout is implemented with Bootstrap 4.

Map is implemented using Leaflet.

API communication is done with Axios.

Refs HP-34


dashboard/src/actions.ts
```diff
@@ -0,0 +1,145 @@
+import { Moment } from 'moment';
+
+import { RegionList, RegionStatsList } from './api/types';
+import { MapViewport } from './components/types';
+
+interface CheckExistingLoginAction {
+    type: 'CHECK_EXISTING_LOGIN';
```
frwickst: This looks really error prone, using the same string in two places instead of using a constant. Now if the interface changes, the function below needs to be updated as well. And there is most likely a reducer catching this as well and it needs to keep up to date with the string.

dashboard/src/api/auth-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+const tokenStorage = new TokenStorage();
+
+export default class AuthManager {
+    /** Maximum age of the token in seconds before it is refreshed. */
+    public maxTokenAge: number = 5 * 60 * 1000;  // 5 minutes
+
+    private _api: Api;
```
frwickst: Style bikeshedding suggestion:
In JS it is usually regarded as bad to add underscore to variable names, and MS TypeScript guide seems to agree regarding TS [1] as well. But it seems to be a bikeshedding issue, for instance the Angular team has discussed this as well[2]. No need to change I guess, just wanted to point it out.

[1] https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names
[2] https://github.com/angular/angular.io/issues/1108

dashboard/src/api/auth-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+        });
+    }
+
+    initiateLogin(username: string, password: string): Promise<CodeToken> {
+        return this._axios.post<CodeToken>(
+            this._api.endpoints.authCodeToken, {username, password}).then(
+                (response) => response.data);
+    }
+
+    continueLogin(
```
frwickst: Style bikeshedding suggestion:
This function declaration is on two lines but the one before it on one line. You probably have a hard limit on 79 chars so I guess that is the reason but it really hurts readability imho.

dashboard/src/api/auth-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+        return this.refreshLogin().then((authToken) => {
+            if (authToken) {
+                this._mountAxiosInterceptor();
+            }
+            return authToken;
+        });
+    }
+
+    initiateLogin(username: string, password: string): Promise<CodeToken> {
+        return this._axios.post<CodeToken>(
```
frwickst: Style bikeshedding suggestion:

IMHO this a a lot less readable then:
```
    initiateLogin(username: string, password: string): Promise<CodeToken> {
        return this._axios.post<CodeToken>(
            this._api.endpoints.authCodeToken, {username, password}
        ).then(
            (response) => response.data
        );
    }
```

dashboard/src/api/auth-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+            }
+        };
+    }
+
+    private _refreshToken: (() => Promise<AuthToken>) = () => {
+        if (this._tokenRefreshPromise) {
+            return this._tokenRefreshPromise;
+        }
+        const refreshUrl = this._api.endpoints.authRefresh;
+        const promise = this._tokenRefreshPromise = this._axios.post(
```
frwickst: Reassignment to unneeded variable, a `return this._tokenRef...` would be enough here.

dashboard/src/api/auth-manager.ts
```diff
@@ -0,0 +1,138 @@
...
+    private _authInterceptorId: number | null = null;
+    private _tokenRefreshPromise: Promise<AuthToken> | null = null;
+
+    constructor(api: Api, axiosInstance: axios.AxiosInstance) {
+        this._api = api;
+        this._axios = axiosInstance;
+    }
+
+    checkExistingLogin(): Promise<AuthToken|null> {
+        return this.refreshLogin().then((authToken) => {
```
frwickst: Most if not all of the promises that I could find the in code has no return type (aka `any`)  when it comes to their return values. Might be something to consider adding.

dashboard/src/api/index.ts
```diff
@@ -0,0 +1,73 @@
+import * as axios from 'axios';
+import { Moment } from 'moment';
+
+import AuthManager from './auth-manager';
+import { RegionList, RegionStatsList } from './types';
+
+interface SuccessCallback<T> {
+    (response: axios.AxiosResponse<T>): void;
```
frwickst: I guess these will be extended later? Otherwise the arguments for both Success and error handlers are unused.

dashboard/src/api/types.ts
```diff
@@ -0,0 +1,68 @@
...
+    type: 'FeatureCollection';
+    features: Array<ApiFeature<I, P>>;
+}
+
+//////////////////////////////////////////////////////////////////////
+// Exported API interfaces
+
+export type Region = ApiFeature<RegionId, RegionProperties>;
+
+export interface RegionList extends
```
frwickst: This looks really unreadable, no idea if it just is just me not knowing TS as well or what but the last {} for instance looks really wierd.

dashboard/src/components/TimeSelect.tsx
```diff
@@ -0,0 +1,203 @@
...
+        strictParsing: true,
+        closeOnSelect: false,
+        closeOnTab: true,
+        utc: false,
+
+        onAutoUpdateChange: (newValue: boolean) => undefined,
+        autoUpdate: false,
+    };
+
+    /*constructor(props: Props) {
```
frwickst: Remove commented out code

dashboard/src/components/ParkingRegionsMap.css
```diff
@@ -0,0 +1,5 @@
+.leaflet-container {
+    min-height: 10rem;
+    //width: 100%;
```
frwickst: Remove commented out code

dashboard/src/components/ParkingRegionsMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+                />
+                {geoJsonElement}
+            </Map>);
+    }
+
+    private bindPopupToRegion = (
+        region: Region,
+        layer: Leaflet.Layer
+    ): void => {
+        const p = region.properties;
```
frwickst: Style bikeshedding suggestion:
I'm usually not a big fan of using single letter variables in cases where readability suffers a lot. In this case, when `p` is used in many places and not just in a single loop or such, it hurts readability because the reader has no idea wher `p` is pointing to.

dashboard/src/components/ParkingRegionsMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+            <b>Paikkoja per km<sup>2</sup></b>:
+              ${p.spotsPerKm2.toFixed(2)}<br>
+            <hr>
+            <b>Pysäköintejä</b>:
+              ${p.parkingCount || 0}<br>
+            <b>Pysäköintejä per km<sup>2</sup></b>:
+              ${parkingsPerKm2.toFixed(2)}<br>
+            <b>Käyttöaste</b>:
+              ${usagePercentage.toFixed(2)}&nbsp;%<br>
+            `;  /* + '<pre>' + JSON.stringify(p, null, 2) + '</pre>'*/
```
frwickst: Remove commented out code.

dashboard/src/components/ParkingRegionsMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+            lengthSum += Object.keys(region.properties).length || 0;
+            lengthSum += region.properties.parkingCount || 0;
+            if (region.properties.isSelected) {
+                selectedId = region.id;
+            }
+        }
+    }
+    const lastRegion = regions[regions.length - 1];
+    return (
+        regions.length.toString() + '/'
```
frwickst: Consider using ` for string/variable concatenation, in this case:
```
`${regions.length.toString()}/${lengthSum}/${lastRegion.id}/${selectedId }`
```

dashboard/src/components/ParkingRegionsMap.tsx
```diff
@@ -0,0 +1,204 @@
...
+}
+
+function getUsageFactorForRegion(
+    p?: RegionProperties | null
+): number | undefined {
+    if (!p || !p.parkingCount || !p.areaKm2) {
+        return undefined;
+    }
+
+    // const capacity = p.capacityEstimate || (p.areaKm2 * 1000);
```
frwickst: Remove commented out code

dashboard/src/components/TimeSelect.tsx
```diff
@@ -0,0 +1,203 @@
...
+    }*/
+
+    getInitialState() {
+        return {
+            ...ReactDatetimeProto.getInitialState.call(this),
+            autoUpdate: AutoUpdatingDatetime.defaultProps.autoUpdate,
+        };
+    }
+
+    getStateFromProps = (props: Props) => {
```
frwickst: Style bikeshedding suggestion:
In the case when using the shorthand notation/new function notation and only returning one value straight away then an even shorter notation can be used:

```
    getStateFromProps = (props: Props) => (
        this.getStateFromPropsOverride(props)
    )
```

dashboard/src/components/TimeSelect.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    <InputGroupButton>
+                        <Button
+                            onClick={this.handleUpdateButtonClick}
+                            className="update-button"
+                            color="info"
+                            outline={true}
+                            active={(this.state as {autoUpdate?: boolean}).autoUpdate}
+                        >
+                            <i className={iconClass}/>
+                            <span className="text"> Pidä ajan tasalla</span>
```
frwickst: Space before text, is that on purpose? If yes, consider using CSS.

dashboard/src/containers/Dashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+        this.timer = null;
+    }
+
+    autoUpdate() {
+        if (this.props.onUpdate) {
+            this.props.onUpdate();
+        }
+    }
+
+    render() {// xl lg md sm xs
```
frwickst: I found this comment not helping me that much, I had no idea why the sizes were listed here since that is not the case in other render methods that has been written.

dashboard/src/containers/Dashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+                            </Card>
+                        </Col>
+                        <Col xl="5" lg="6" md="6" sm="12">
+                            <Card>
+                                <CardBody>
+                                    <RegionSelector/>
+                                </CardBody>
+                            </Card>
+                            <Card className="map-card">
+                                {/*<CardHeader>Kartta</CardHeader>*/}
```
frwickst: Remove commented out code

dashboard/src/containers/Dashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+                            </Card>
+                        </Col>
+                    </Row>
+                    <Row>
+                        <Col>
+                            <Button
+                                onClick={this.props.onLogout}
+                                color="danger"
+                            >
+                                <i className="fa fa-sign-out"/>{' '}Kirjaudu ulos
```
frwickst: Umm `{' '}`? Is this for styling? If yes, consider CSS.

dashboard/src/containers/TimeSelect.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import TimeSelect, { Props } from '../components/TimeSelect';
+import * as dispatchers from '../dispatchers';
+import { RootState } from '../types';
+
+function mapStateToProps(state: RootState): Partial<Props> {
+    return {
+        value: (state.dataTime) ? moment(state.dataTime) : undefined,
+        autoUpdate: state.autoUpdate,
+        // closeOnTab: false,
```
frwickst: Remove commented out code.

dashboard/src/reducers.ts
```diff
@@ -0,0 +1,181 @@
...
+const initialAuthState: AuthenticationState = {
+    loggedIn: false,
+    existingLoginChecked: false,
+};
+
+function auth(
+    state: AuthenticationState = initialAuthState,
+    action: Action,
+): AuthenticationState {
+    if (action.type === 'CHECK_EXISTING_LOGIN') {
```
frwickst: As was commented above, concider using constants.

dashboard/src/actions.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+export function checkExistingLogin(): CheckExistingLoginAction {
```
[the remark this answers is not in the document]
suutari-ai: Shorthand? It's possible to use lambdas too, but IMHO regular functions look nicer and I like to use them when possible. (The not-possible case is usually because the different this binding of lambdas makes it wiser to use lambdas for some this-using class methods.)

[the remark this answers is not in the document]
frwickst: > Shorthand?
I find the notication to be shorter and in most cases better then the `function ...` annotatation, but this should have been marked as bikeshed topic as well :)

dashboard/src/actions.ts
suutari-ai: The reducer won't work if it's using a different string, because then the type check will fail. I know this could use some refactoring so that the same wouldn't have to be written twice, but didn't do that yet, because it didn't occur to me right away how to do it properly (and wanted to have this PR:ed before...)

dashboard/src/api/auth-manager.ts
suutari-ai: So, what's your suggestion for private variable naming?

frwickst: `private api: Api;` ?

dashboard/src/api/auth-manager.ts
suutari-ai: Long lines hurt the readability more IMO 😄

dashboard/src/actions.ts
frwickst: Of course the reducer won't work, that is point 😄 To make it harder to make the mistake. But as you said, this can be refactored later if needed.

dashboard/src/api/auth-manager.ts
frwickst: So does inconsistency 😉 But again, bikeshedding, no need for changes.

dashboard/src/api/auth-manager.ts
suutari-ai: This is because I thought that `this._tokenRef..` could be changed to null while prosessing the promise and I wasn't sure if it's possible that the Promise is already resolved (which would set it to null) at line 136 which returns the promise.

frwickst: Alright... I'll allow, but this is a bit hard to understand and the reassignment does not help here.

dashboard/src/api/index.ts
suutari-ai: It's a callback interface. It means that the you give it a function that will be called with a response. It's not unused, see line 61 where callback is called with a response.

frwickst: Alright 👍

dashboard/src/api/types.ts
suutari-ai: Are you referring to this? (Can't be sure, since the comment wasn't on the last line as I would expect it be on...)

```typescript
export interface RegionList extends
ApiFeatureCollection<RegionId, RegionProperties>,
PaginatedList {
}
```

It's declaring an interface which extends two other interfaces, but not adding any new properties. Don't know how to make it more readable...

dashboard/src/components/ParkingRegionsMap.css
suutari-ai: OK, will do

dashboard/src/components/ParkingRegionsMap.tsx
suutari-ai: Good suggestion. Thanks!

dashboard/src/api/auth-manager.ts
suutari-ai: Do you mean the authToken being without a type definition in this line?

dashboard/src/api/auth-manager.ts
suutari-ai: I don't really know which I like more... maybe I gotta revisit my styling practices for TS code, but won't change this now, since there would be many other similar places to change and I'm not yet even sure which is the best style. :)

dashboard/src/components/TimeSelect.tsx
suutari-ai: OK

dashboard/src/components/TimeSelect.tsx
suutari-ai: Yes it's on purpose. Don't know how to do that with CSS.

dashboard/src/containers/Dashboard.tsx
suutari-ai: Oops.. it was left there by accident.. were there just to remind me about the bootstrap classes which I'm not that familiar with yet... Will remove.

dashboard/src/containers/Dashboard.tsx
suutari-ai: Yeah, same here. Don't know CSS that well...

dashboard/src/containers/Dashboard.tsx
suutari-ai: OK

dashboard/src/reducers.ts
suutari-ai: That would only add more code without actually improving anything.

Tried to change that string on the line to `'CHECK_EXISTING_LOGIN2'` and this is what TypeScript says after that:
```
/home/tuomas/projects/hel/parkkihubi/app/dashboard/src/reducers.ts
(21,9): Operator '===' cannot be applied to types '"CHECK_EXISTING_LOGIN" | "RESOLVE_EXISTING_LOGIN_CHECK" | "REQUEST_CODE_TOKEN" | "RECEIVE_CODE_TO...' and '"CHECK_EXISTING_LOGIN2"'.
```

frwickst: Ok

dashboard/src/containers/Dashboard.tsx
frwickst: Ok

dashboard/src/components/TimeSelect.tsx
frwickst: Ok

dashboard/src/api/auth-manager.ts
frwickst: Yes 🙂

dashboard/src/actions.ts
suutari-ai: sorry, I meant won't compile. I.e. you'll spot that error anyway

dashboard/src/api/types.ts
frwickst: I refered to the entire export statement, but excecially to the last {}. Can't it be written as
```
export interface RegionList extends
ApiFeatureCollection<RegionId, RegionProperties>,
PaginatedList {}
```
for instance?

dashboard/src/api/auth-manager.ts
frwickst: Ok

dashboard/src/api/auth-manager.ts
suutari-ai: It's not actually `any`, since that's not even allowed (`"noImplicitAny": true,` in `tsconfig.json`). TypeScript just knows the type from the context.

PR #57, opened 2018-02-19
Dashboard: Implement active parkings table

suutari-ai:
Add monitoring API endpoint for listing valid parkings at given time and
use that to list the active parkings at selected time and region in the
Dashboard.

Refs HP-35

### DIFF
--- a/parkings/api/monitoring/serializers.py
+++ b/parkings/api/monitoring/serializers.py
@@ -1,0 +1,26 @@
+import rest_framework_gis.serializers as gis_serializers
+from rest_framework import serializers
+
+from ...models import Parking
+
+
+class ParkingSerializer(gis_serializers.GeoFeatureModelSerializer):
+    operator_name = serializers.CharField(
+        source='operator.name', read_only=True)
+
+    class Meta:
+        model = Parking
+        fields = [
+            'id',
+            'registration_number',
+            'location',
+            'region',
+            'zone',
+            'terminal_number',
+            'operator_name',
+            'time_start',
+            'time_end',
+            'created_at',
+            'modified_at',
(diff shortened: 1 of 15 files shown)


parkings/api/monitoring/serializers.py
```diff
@@ -0,0 +1,27 @@
...
+    operator_name = serializers.CharField(
+        source='operator.name', read_only=True)
+
+    class Meta:
+        model = Parking
+        fields = [
+            'id',
+            'registration_number',
+            'location',
+            # 'parking_area',
```
tuomas777: Is this intentionally left here? And if it is, I don't really see the point.

parkings/api/monitoring/valid_parking.py
```diff
@@ -0,0 +1,38 @@
...
+        ]
+        strict = django_filters.STRICTNESS.RAISE_VALIDATION_ERROR
+
+    def filter_time(self, queryset, name, value):
+        return queryset.valid_at(value)
+
+
+class ValidParkingViewSet(viewsets.ReadOnlyModelViewSet):
+    permission_classes = [MonitoringApiPermission]
+    queryset = Parking.objects.all().order_by('time_start')
```
tuomas777: This could have `select_related('operator')` as the serializer is using `operator.name` ?

suutari-ai: Good point. Will add.

parkings/api/monitoring/serializers.py
suutari-ai: Yeah, there's no much use having it there as a comment. I will remove it.

PR #59, opened 2018-08-22
Travis: Update PostgreSQL to 10

suutari-ai:
Use PostgreSQL 10 to test, because that is in use on the servers.

Also make Python 3.5 the default Python version on tests, because that
is in use in the servers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,11 @@ matrix:
     - python: "3.5"
     - python: "3.6"
     - env: TOXENV=requirements
-      python: "3.4"
+      python: "3.5"
       addons: null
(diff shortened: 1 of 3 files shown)


suutari-ai:
Hmm.. can't find a reason why Python 3.4 installation fails on Travis. This is reported on its logs:
```
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/16.04/x86_64/python-3.4.tar.bz2
$ curl -sSf -o python-3.4.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 403 Forbidden
Unable to download 3.4 archive. The archive may not exist. Please consider a different version.
```

PR #64, opened 2019-05-23
Add a better importer to import PermitAreas

urshala:
Add a better importer to import PermitAreas

Refs HP-61

### DIFF
--- a/parkings/importers/__init__.py
+++ b/parkings/importers/__init__.py
@@ -1,1 +1,7 @@
-from .parking_areas import ParkingAreaImporter  # noqa
+from .parking_areas import ParkingAreaImporter
+from .permit_areas import PermitAreaImporter
+
+__all__ = [
+    'ParkingAreaImporter',
+    'PermitAreaImporter'
+]
--- a/parkings/importers/permit_areas.py
+++ b/parkings/importers/permit_areas.py
@@ -1,0 +1,99 @@
+import logging
+
+from django.db import transaction
+from lxml import etree
+from owslib.wfs import WebFeatureService
+
+from parkings.models import PermitArea
+
+from .utils import get_polygons
+
+logger = logging.getLogger(__name__)
+
+
+class PermitAreaImporter(object):
+
+    """
+    Imports permit area data from kartta.hel.fi.
+    """
+
+    def __init__(self, overwrite=False):
+        self.ns = {
+            'wfs': 'http://www.opengis.net/wfs/2.0',
+            'avoindata': 'https://www.hel.fi/avoindata',
+            'gml': 'http://www.opengis.net/gml/3.2',
+        }
+        self.created = 0
+
+    def import_permit_areas(self):
+        response = self._download()
+        permit_areas_dict = None
+
+        if response is not None:
+            permit_areas_dict = self._parse_permit_areas(response)
+        else:
+            logger.error("Download failed.")
+            return False
+
+        if permit_areas_dict is not None:
+            self._save_permit_areas(permit_areas_dict)
+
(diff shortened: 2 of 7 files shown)


parkings/importers/permit_areas.py
```diff
@@ -0,0 +1,110 @@
...
+            'gml': 'http://www.opengis.net/gml/3.2',
+        }
+        self.created = 0
+
+    def import_permit_areas(self):
+        response = self._download()
+        permit_areas_dict = None
+
+        if response is not None:
+            PermitArea.objects.all().delete()
```
suutari-ai: Rather than deleting them all, create a list of the ids that were imported and delete everything else. This way it would stay as atomic operation and wouldn't clear anything if there's an error.

parkings/importers/permit_areas.py
```diff
@@ -0,0 +1,110 @@
...
+        if permit_areas_dict is not None:
+            self._save_permit_areas(permit_areas_dict)
+
+    @transaction.atomic
+    def _save_permit_areas(self, permit_areas_dict):
+        logger.info('Saving permit areas.')
+
+        for index, permit_area_dict in enumerate(permit_areas_dict):
+            permit_area = PermitArea(**permit_area_dict)
+            permit_area.save()
```
suutari-ai: Do it like this instead
 ```python
area_ids = []
for area_dict in permit_area_dicts:
    (area, created, updated) = create_or_update(
        PermitArea, 'identifier', permit_area_dict)
    area_ids.append(area.pk)
PermitArea.objects.exclude(pk__in=area_ids).delete()
```
with
```python
def create_or_update(model, key_field, data):
    values = data.copy()
    key_lookup = {key_field: values.pop(key_field)}
    (obj, created) = model.objects.get_or_create(defaults=values, **key_lookup)
    changed = False
    if not created:
        for (key, value) in values.items():
             if getattr(obj, key) != value:
                setattr(obj, key, value)
                changed = True
        if changed:
            obj.save()
    return (obj, created, changed)
```

parkings/importers/permit_areas.py
```diff
@@ -0,0 +1,110 @@
...
+            # get leaf nodes. Treat LinearRing and MultiSurface the same way
+            if len(pos) == 0:
+                positions = list(filter(None, pos.text.split(' ')))
+                points = []
+                points_as_pairs = zip(positions[1::2], positions[::2])
+                for latitude, longitude in points_as_pairs:
+                    points.append(Point(float(latitude), float(longitude)))
+                polygons.append(Polygon(points))
+
+        return MultiPolygon(polygons)
```
suutari-ai: Is this function copied from another importer? If it is, you should rather move it to a common place so that it can be reused rather than copy-pasted.

parkings/importers/__init__.py
```diff
@@ -1 +1,2 @@
 from .parking_areas import ParkingAreaImporter  # noqa
+from .permit_areas import PermitAreaImporter  # noqa
```
suutari-ai: The correct way to do this is: ```
from .parking_areas import ParkingAreaImporter
from .permit_areas import PermitAreaImporter
__all__ = [
    'ParkingAreaImporter',
    'PermitAreaImporter',
]
```
so you don't need to add these nasty `# noqa` comments.

suutari-ai:
Looks very good, Deepak! Thanks a lot!

There's some minor details I'd like be improved though. Added those as inline comments. Please check them out.

suutari-ai:
Thanks Deepak! I rebased the commits once more and merged them to the scanner-car branch (which is heading to master in PR #66).

PR #65, opened 2019-05-24
Add an API endpoint for checking parking validity

japauliina:
This endpoint is meant for scanner car to check if parking of a vehicle is valid or not. Both parkings and permits are checked.


suutari-ai:
This is already incorporated into #69 and #82.

PR #70, opened 2019-06-12
Fix PermitCacheItem not being saved

japauliina:
PermitCacheItems were not saved when Permits were created via API in bulks.


parkings/tests/api/enforcement/test_permit.py
```diff
@@ -104,3 +104,26 @@ def test_permit_data_matches_permit_object(staff_api_client, permit):
...
+    permit_data = [
+        {
+            "series": permit_series.id,
+            "external_id": generate_external_ids(),
+            "subjects": generate_subjects(),
+            "areas": generate_areas(),
+        },
+        {
+            "series": permit_series.id,
+            "external_id": generate_external_ids(),
```
suutari-ai: This could conflict with the previously created external ID. Use static values instead to avoid such conflicts, e.g. `"E1"` for the first external_id and `"E2"` for the second.

PR #72, opened 2019-06-13
Delete old PermitSeries when activating new series

japauliina:
Clean older PermitSeries (+ Permits and PermitCacheItems belonging to the series) when activating a new series to keep their count minimal.


parkings/models/permit.py
```diff
@@ -30,6 +34,12 @@ def active(self):
     def latest_active(self):
         return self.active().order_by('-modified_at').first()
 
+    def prunable(self):
+        pruneable_after_date = timezone.now() - datetime.timedelta(
+            days=settings.PARKKIHUBI_PERMITS_PRUNABLE_AFTER_DAYS
+        )
+        return self.filter(created_at__lt=pruneable_after_date)
```
suutari-ai: Maybe add `active=False` to this?

parkings/tests/api/enforcement/test_activate_permit_series.py
```diff
@@ -0,0 +1,26 @@
...
+    old_permit_series = PermitSeries.objects.create(active=True)
+    old_permit_series.created_at = timezone.now() - datetime.timedelta(3)
+    old_permit_series.save()
+
+    assert PermitSeries.objects.all().count() == 2
+    response = staff_api_client.post(get_url(new_permit_series) + 'activate/')
+    assert response.data['status'] == "OK"
+    assert response.status_code == HTTP_200_OK
+    assert PermitSeries.objects.latest_active() == new_permit_series
+    assert PermitSeries.objects.all().count() == 1
```
suutari-ai: Maybe this could be split to two test cases instead? (one checking that the activation works, i.e. deactivates other series and activates the one and another test that checks that the old series are deleted)

parkkihubi/settings.py
```diff
@@ -212,3 +212,4 @@
 PARKKIHUBI_OPERATOR_API_ENABLED = env.bool('PARKKIHUBI_OPERATOR_API_ENABLED', True)
 PARKKIHUBI_ENFORCEMENT_API_ENABLED = (
     env.bool('PARKKIHUBI_ENFORCEMENT_API_ENABLED', True))
+PARKKIHUBI_PERMITS_PRUNABLE_AFTER_DAYS = 3
```
suutari-ai: This could use timedelta like the other duration settings, e.g.
```python
PARKKIHUBI_PERMITS_PRUNABLE_AFTER = timedelta(days=3)
```

parkings/tests/api/enforcement/test_activate_permit_series.py
suutari-ai: ...and since `activate_permit_series` is already part of the test filename, the function names don't have to have that as part of their name. The tests could be named e.g. `test_series_gets_activated` and `test_old_series_are_pruned`.

PR #73, opened 2019-06-17
Refactor the importers 

urshala:
* Refactor the importer so that duplication of codes is abstracted
* Add a better importer for paymentZones

Refs HP-61

### DIFF
--- a/parkings/importers/payment_zones.py
+++ b/parkings/importers/payment_zones.py
@@ -1,0 +1,64 @@
+import logging
+
+from django.db import transaction
+
+from parkings.models import PaymentZone
+
+from .utils import BaseImporterMixin
+
+logger = logging.getLogger(__name__)
+
+
+class PaymentZoneImporter(BaseImporterMixin):
+
+    """
+    Imports paymentzones data from kartta.hel.fi.
+    """
+
+    def __init__(self, overwrite=False):
+        super().__init__(overwrite=overwrite)
+        self.typename = 'Pysakoinnin_maksuvyohykkeet_alue'
+
+    def import_payment_zones(self):
+        response = self._download(self.typename)
+        payment_zones_dict = None
+
+        if response is not None:
+            payment_zones_dict = self._parse_data(response)
+        else:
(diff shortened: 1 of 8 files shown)


parkings/importers/payment_zones.py
```diff
@@ -0,0 +1,96 @@
...
+from owslib.wfs import WebFeatureService
+
+from parkings.models import PaymentZone
+
+from .utils import get_polygons
+
+logger = logging.getLogger(__name__)
+
+
+class PaymentZoneImporter(object):
```
japauliina: Quickly looking this seems to have quite a lot of duplication with `ParkingAreaImporter`. I wonder if something should be done about it or will the importers just get overly complicated... Would like to hear @suutari-ai  opinion about this.

parkings/tests/payment_zone_importer_data.xml
```diff
@@ -0,0 +1,19 @@
...
+            <gml:LinearRing>
+              <gml:posList>2.5497165701059E7 6673221.131005 2.5497167810273E7 6673221.211415 2.5497167006736E7 6673233.337525 2.5497165052488E7 6673233.1924 2.5497165701059E7 6673221.131005</gml:posList>
+            </gml:LinearRing>
+          </gml:exterior>
+        </gml:Polygon>
+      </avoindata:geom>
+    </avoindata:Pysakoinnin_maksuvyohykkeet_alue>
+  </wfs:member>
+</wfs:FeatureCollection>
+
```
japauliina: This empty line can be deleted I think :)

parkings/importers/payment_zones.py
suutari-ai: I think you're right about the duplication. IMHO such duplication should be avoided and rather the importers should be refactored so that they can share as much of the code as possible.

urshala: That is a very good point. Indeed this is almost the exact copy of `ParkingAreaImporter`. I will refactor to abstract the duplication. Thank you @japauliina , @suutari-ai .

PR #83, opened 2019-07-14
Upgrade Django to 2.2.3

suutari-ai:
Make code compatible with Django 2.2 and upgrade Django to the latest
version, i.e. 2.2.3.


suutari-ai:
Note: This is based on PR #82 and so all commits from it are included too. Review will be easier after PR #82 is merged and those commits won't be visible in this PR anymore.

PR #93, opened 2019-11-19
Disc parking

urshala:
* Modify the parking model to add `is_disc_parking` field to specify whether a parking is normal or disc. 
  Make zone optional field in the model.
* Change mandatory fields based on the parking type.
* Add tests for disc parking


parkings/api/operator/parking.py
```diff
@@ -30,12 +30,22 @@ class Meta:
...
         self.fields['time_end'].timezone = pytz.utc
+        self.fields['zone'].required = True
+
+        self._set_required_extra_fields()
+
+    def _set_required_extra_fields(self):
+        # if it is disc parking, make location required and zone optional field.
+        if getattr(self, 'initial_data', None) and self.initial_data.get('is_disc_parking', False):
+            self.fields['location'].required = True
+            self.fields['zone'].required = False
```
suutari-ai: Use a separate variable for the getattr result to make this code cleaner:
```python
initial_data = getattr(self, 'initial_data', None)
if initial_data and initial_data.get('is_disc_parking'):
```

Also, remove the comment, as it's saying just the same thing as can be read from the code.

PR #108, opened 2020-01-14
Enforcement disc parking

urshala:
* Add disc parking to enforcement endpoint
* Replace empty `time_end` value by hard coded date-time string


parkings/tests/api/enforcement/test_valid_parking.py
```diff
@@ -97,6 +114,9 @@ def check_parking_data_matches_parking_object(parking_data, parking_obj):
     assert parking_data['time_end'] == iso8601(parking_obj.time_end)
     assert parking_data['operator_name'] == str(parking_obj.operator.name)
 
+    if parking_obj.is_disc_parking:
+        assert parking_data['is_disc_parking'] == parking_obj.is_disc_parking
+
```
suutari-ai: maybe this should check the value in both cases (paid or disc), i.e.:
```python
assert parking_data.get('is_disc_parking', False) == parking_obj.is_disc_parking
```

parkkihubi/settings.py
```diff
@@ -214,3 +214,5 @@
 PARKKIHUBI_ENFORCEMENT_API_ENABLED = (
     env.bool('PARKKIHUBI_ENFORCEMENT_API_ENABLED', True))
 PARKKIHUBI_PERMITS_PRUNABLE_AFTER = timedelta(days=3)
+PARKKIHUBI_NONE_END_TIME_REPLACEMENT = env.str(
+    'PARKKIHUBI_NONE_END_TIME_REPLACEMENT', default='2030-12-31T23:59:59Z')
```
suutari-ai: Move this new setting to line 210 just before all the API_ENABLED settings. That's more logical place for it, since it's related to the times.

suutari-ai:
About the second commit message: The system is called PASI (not PASSI). Also don't mention the system name in the commit message title, since the feature is not only for single system. The title could be e.g. "Make it possible to override empty end times in valid_parking" and then in the commit message body you could tell that e.g. PASI has a problem with null values in `time_end` field and that's the reason for this new feature.

parkings/api/enforcement/valid_parking.py
```diff
@@ -33,6 +33,11 @@ def to_representation(self, instance):
         if not instance.is_disc_parking:
             representation.pop('is_disc_parking')
 
+        if instance.time_end is None:
+            replacement_value = getattr(
+                settings, 'PARKKIHUBI_NONE_END_TIME_REPLACEMENT', None)
+            representation['time_end'] = replacement_value
```
suutari-ai: add `or None`  to this so that it is possible to use empty string in the setting to get this set to None

suutari-ai: since the setting could be set via env variable, and you can't pass None from there

suutari-ai:
Good work Deepak! :+1:

PR #109, opened 2020-01-15
Enforcement domain model

amanpdyadav:
Added enforcement domain model and its relation to Parking,
ParkingArea, ParkingTerminal, PermitArea, Permit and PaymentZone.
Migrations to handle the updates of tables.


parkings/models/enforcement_domain.py
```diff
@@ -0,0 +1,24 @@
...
+from django.contrib.gis.db import models as gis_models
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from .constants import GK25FIN_SRID
+from .mixins import TimestampedModelMixin
+
+
+class EnforcementDomain(TimestampedModelMixin):
+    enforcers = models.ManyToManyField(User, verbose_name=_("enforcers"))
```
suutari-ai: Use `settings.AUTH_USER_MODEL` here.

parkings/admin.py
```diff
@@ -3,11 +3,17 @@
...
 from .admin_utils import ReadOnlyAdmin, WithAreaField
 from .models import (
-    EnforcementDomain, Operator, Parking, ParkingArea, ParkingCheck,
+    EnforcementDomain, Enforcer, Operator, Parking, ParkingArea, ParkingCheck,
     ParkingTerminal, PaymentZone, Permit, PermitArea, PermitLookupItem,
     PermitSeries, Region)
 
 
+@admin.register(Enforcer)
+class Enforcer(WithAreaField, OSMGeoAdmin):
```
suutari-ai: the admin class should be named `EnforcerAdmin`

parkings/migrations/0029_enforcement_domain.py
```diff
@@ -14,16 +15,57 @@ class Migration(migrations.Migration):
...
+        migrations.AddField(
+            model_name='permitseries',
+            name='owner_enforcer',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, related_name='permit_series_set', to='parkings.Enforcer'),
+        ),
+        migrations.AlterField(
+            model_name='paymentzone',
+            name='code',
+            field=models.CharField(max_length=10),
+        ),
```
suutari-ai: Move these changes to the `code` field up to the `AddField`-statement of the `code` field to avoid adding the field and changing it in same migration.

parkings/models/enforcement_domain.py
```diff
@@ -10,8 +10,7 @@
 class EnforcementDomain(TimestampedModelMixin):
     code = models.CharField(max_length=10, unique=True, verbose_name=_('code'))
     name = models.CharField(max_length=40, verbose_name=_('name'))
-    geom = gis_models.MultiPolygonField(
-        srid=GK25FIN_SRID, verbose_name=_('geometry'))
+    geom = gis_models.MultiPolygonField(srid=GK25FIN_SRID, verbose_name=_('geometry'), default="MULTIPOLYGON EMPTY")
```
suutari-ai: Maybe the geom field should be nullable?

Then we could easily check, if the geometry of the ED is specified or not. That could affect some business logic later.

parkings/migrations/0029_enforcement_domain.py
```diff
@@ -14,16 +15,57 @@ class Migration(migrations.Migration):
...
+            },
+        ),
+        migrations.CreateModel(
+            name='Enforcer',
+            fields=[
+                ('created_at', models.DateTimeField(auto_now_add=True, verbose_name='time created')),
+                ('modified_at', models.DateTimeField(auto_now=True, verbose_name='time modified')),
+                ('id', models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False)),
+                ('name', models.CharField(max_length=80, verbose_name='name')),
+                ('enforced_domain', models.ForeignKey(help_text='An enforcement domain than is enforced by this enforcer', null=True, on_delete=django.db.models.deletion.PROTECT, to='parkings.EnforcementDomain', verbose_name='enforced domain')),
```
suutari-ai: There seems to be a typo in this help_text (probably written by me :smiley_cat:). The `than` word should be `that`, of course.

parkings/models/enforcement_domain.py
amanpdyadav: If I put **null=True**. It breaks the admin view. So had to set **default="MULTIPOLYGON EMPTY"** instead of **null=True**

parkings/migrations/0029_enforcement_domain.py
suutari-ai: Remove the `null=True, blank=True` part, since every enforcer should have an ED.

And since there always is a single enforcement domain for each Enforcer, the `An` article in the `help_text` should be replaced with `The`, i.e. the whole help text should be: "The enforcement domain enforced by this enforcer".

parkings/admin.py
```diff
@@ -3,8 +3,21 @@
...
+
+
+@admin.register(Enforcer)
+class EnforcerAdmin(WithAreaField, OSMGeoAdmin):
+    list_display = ['id', 'name', 'users', 'enforced_domain']
+    ordering = ('name',)
+
+
+@admin.register(EnforcementDomain)
+class EnforcmenteDomainAdmin(WithAreaField, OSMGeoAdmin):
```
suutari-ai: Enforcmente -> Enforcement

parkings/models/enforcement_domain.py
```diff
@@ -0,0 +1,44 @@
...
+            code=code, defaults={"name": name})[0]
+
+
+class Enforcer(TimestampedModelMixin, UUIDPrimaryKeyMixin):
+    name = models.CharField(verbose_name=_("name"), max_length=80)
+    users = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.PROTECT, verbose_name=_("users"))
+    enforced_domain = models.ForeignKey(
+        EnforcementDomain, verbose_name=_("enforced domain"), on_delete=models.PROTECT,
+        help_text=_("An enforcement domain that is enforced by this enforcer"))
```
suutari-ai: "The enforcement domain enforced by this enforcer"

parkings/models/enforcement_domain.py
```diff
@@ -0,0 +1,44 @@
...
+    def get_default_domain(cls):
+        (name, code) = settings.DEFAULT_ENFORCEMENT_DOMAIN
+        return cls.objects.get_or_create(
+            code=code, defaults={"name": name})[0]
+
+
+class Enforcer(TimestampedModelMixin, UUIDPrimaryKeyMixin):
+    name = models.CharField(verbose_name=_("name"), max_length=80)
+    users = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.PROTECT, verbose_name=_("users"))
```
suutari-ai: Should probably be OneToOne field called "user" and then the PermitSeries owner could be implemented as a FK to User rather than two separate FKs (owner_operator and owner_enforcer).

suutari-ai:
Replaced with #115 to get a proper codecov report.

PR #111, opened 2020-01-23
Parking forecast: Add ActiveParking cache model

klemmari1:
Add parking count cache model for regions and parking areas. This makes it easier to gather data for the forecast model and also it can be used to get precalculated parking count results for the API.

- Add script for caching parking counts for regions and parking areas.
- Add management commands to run the caching scripts.


parkings/forecasts/cache_parking_counts.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python
```
suutari-ai: This file doesn't have a `if __name__ == '__main__'` part so there's no need to have a shebang either. Please remove it.

parkings/management/commands/cache_parking_counts_by_parking_area.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
```
suutari-ai: Same here, remove the shebang.

parkings/management/commands/cache_parking_counts_by_region.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
```
suutari-ai: Remove the shebang.

suutari-ai:
Looks very good. Nice work!

The code test coverage is pretty low though: Codecov reported "The diff coverage is 64.28%". Can you add some tests to get this number up?

parkings/models/region.py
```diff
@@ -12,11 +12,25 @@
 class RegionQuerySet(models.QuerySet):
     def with_parking_count(self, at_time=None):
         time = at_time if at_time else timezone.now()
-        valid_parkings_q = (
-            Q(parkings__time_start__lte=time) &
-            (Q(parkings__time_end__gte=time) | Q(parkings__time_end=None)))
-        return self.annotate(
-            parking_count=Count(Case(When(valid_parkings_q, then=1))))
```
suutari-ai: Please don't change the formatting of the existing code. It seems that you didn't touch the functionality of this code at all.

klemmari1: The point of this change is to use the ActiveParkings model to get the forecasted parking counts for future datetime queries:
```
        if time > timezone.now():
            region_q = self.filter(
                active_parkings__time=time, active_parkings__is_forecast=True
            )
```
or the pre-calculated active parking count for past datetime queries:
```
        else:
            region_q = self.filter(
                active_parkings__time=time, active_parkings__is_forecast=False
            )
```
if there exists an ActiveParking instance for that datetime.

Should I separate this to its own QuerySet method perhaps?

parkings/forecasts/cache_parking_counts.py
```diff
@@ -0,0 +1,85 @@
...
+
+    parkings_end = Parking.objects.all().order_by("-time_start").first().time_end
+
+    current_datetime = datetime.datetime(
+        parkings_start.year,
+        parkings_start.month,
+        parkings_start.day,
+        parkings_start.hour,
+        tzinfo=parkings_start.tzinfo,
+    )
```
suutari-ai: Use `current_datetime = parkings_start.replace(minutes=0, seconds=0, microseconds=0)` here since this could break when there is a DST change. And it's not really "current" datetime, rather `start_hour` or something like that.

suutari-ai:
Couple other things:
 * The many-to-many field to parkings seems to be unused and IMHO it should be anyway removed, since it doesn't probably add any value to have it there (but would rather make it much slower to update those cached instances).
 * The model name is not very descriptive, since "ActiveParking" sounds like those are currently active parkings. How about something like "ParkingCount".

parkings/forecasts/cache_parking_counts.py
suutari-ai: Actually, after checking the code below more carefully, even better variable name would be just `hour`, since it's going to be iterating over all the hours.

klemmari1:
> - The many-to-many field to parkings seems to be unused and IMHO it should be anyway removed, since it doesn't probably add any value to have it there (but would rather make it much slower to update those cached instances). 

Yep, I can remove it. I added it originally to because I was thinking it could be maybe beneficial later if the actual parkings would be cached at the same time, so they could be accessed from the same cache instance, but it is true that it would make the caching much slower.

> - The model name is not very descriptive, since "ActiveParking" sounds like those are currently active parkings. How about something like "ParkingCount".

I initially thought of naming it `ParkingCount` but then I was thinking that is it confusing when accessing the parking count attribute of a parking count object (`parking_count.parking_count`). The reasoning behind the name `ActiveParking` is that you are getting the count of active parkings in region x at time stamp [y, z]. (`ActiveParking.objects.get(region=x, time_start=y, time_end=z).parking_count`)

parkings/forecasts/cache_parking_counts.py
```diff
@@ -0,0 +1,73 @@
...
+def cache_parking_counts_by_region():
+    parkings_start_instance = (
+        ParkingCount.objects.filter(region__isnull=False, is_forecast=False)
+        .order_by("time")
+        .last()
+    )
+    if parkings_start_instance:
+        parkings_start = parkings_start_instance.time
+    else:
+        parkings_start = Parking.objects.all().order_by("time_start").first().time_start
```
suutari-ai: If there is no parkings at all, this will fail, since first return None.

parkings/forecasts/cache_parking_counts.py
```diff
@@ -0,0 +1,73 @@
...
+
+    parkings_end = Parking.objects.all().order_by("-time_start").first().time_end
```
suutari-ai: Sorry, didn't notice this earlier, but there's still a bug in this line, since time_end could be null.

Maybe it should rather use `end_time = timezone.now()`, since it  might be that the parkings end before current time, but we would still want to generate the 0 counts for those hours between the last parking and current time.

PR #116, opened 2020-01-24
Model support for Enforcement Domains

amanpdyadav:
Implement model changes to support multiple cities in the same database
by segmenting their data by so-called enforcement domains.

Add EnforcementDomain model and FKs to it from Parking, ParkingArea,
ParkingTerminal, PermitArea, Permit and PaymentZone.

Add Enforcer model for specifying those Users that are allowed to use
the Enforcement API and FKs from it to EnforcementDomain.

Add three-step migrations for the changes that do the following:

  1. Create EnforcementDomain model and their relation to Parking,
     - ParkingArea, ParkingTerminal, PermitArea, Permit and PaymentZone
       models with null True.
     - Create Enforcer model, add FK to Enforcement domain with default
       null and OneToOne relation to User with null True.
     - Add code and number fields to PaymentZone, owner field to
       PermitSeries with FK to User and an identifier field to
       PermitArea
     - Combination of code and identifier of Permitarea to be unique.

  2. Data migration to create:
     - An EnforcementDomain combining PaymentZones with a union
       operation and assigning it to Parking, ParkingArea,
       ParkingTerminal, PermitArea, Permit and PaymentZone models.
     - An Enforcer for each User with is_staff True and assign it to
       the User.
     - Permit owner with username PASI and assign it to all the
       PermiSeries.

  3. Set following fields to not allow null
     - user of Enforcer model.
     - domain of Parking, ParkingArea, ParkingTerminal, PermitArea,
       Permit and PaymentZone
     - code of PaymentZone.
     - Combination of code and domain of PaymentZone to be unique.

Set default domain to post a request for adding Permit through
enforcement API during the validation of Models in PermitSerializer.

Update test to support model change for multi-city support with
following changes:
  - Add PASI user factory
  - Add enforcement domain to invalid_geom_parking_areas.yaml and
    its id to all the Parking area init.
  - Update test_activate_permit_series PermitSeries to have
    pasi_user as owner.


suutari-ai:
Please rebase the commit on top of the current multicity branch and open a new PR then.

PR #117, opened 2020-01-24
Remove pasi user factory

amanpdyadav:
Pasi user in data migration is staff.

Pasi user factory is not needed as staff user can be an owner of
Permitseries.


parkings/migrations/0030_domain_data.py
```diff
@@ -73,6 +73,7 @@ def _get_or_create_permit_owner(apps):
         username='PASI',
         defaults={
             'is_superuser': True,
+            'is_staff': True
```
suutari-ai: Can you also remove the is_superuser, since it shouldn't be a superuser?

suutari-ai:
Thanks Aman! I'll merge as soon as Travis gives its OK.

PR #118, opened 2020-01-28
Permitted user for PermitArea

amanpdyadav:
Add permitted_user a FK to User table in order to identify the Permit
area the user has access to read.

Update 3 migrations to reflect the change:
1. 0029_enforcement_domain add foreign key to User table from
PermitArea with default null

2. 0030_domain_data update PermitArea's permitted_user with PASI user
as default

3. 0031_default_domain update PermitArea for not to accept default
value as null

Update PermitArea and Geojson importer to pass permitted-user as an
argument.

Update test_geojson_importing_management_commands and
test_importing_management_commands to pass PASI user to be used as
permitted_user for PermitArea

### DIFF
--- a/parkkihubi_hel/importers/permit_areas.py
+++ b/parkkihubi_hel/importers/permit_areas.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.contrib.auth import get_user_model
 from django.db import transaction
 
 from parkings.models import EnforcementDomain, PermitArea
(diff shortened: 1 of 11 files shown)


parkkihubi_hel/importers/permit_areas.py
```diff
@@ -15,20 +16,25 @@ class PermitAreaImporter(WfsImporter):
...
-    def _save_permit_areas(self, permit_areas_dict):
+    def _save_permit_areas(self, permit_areas_dict, permitted_user):
         logger.info('Saving permit areas.')
         count = 0
         permit_area_ids = []
         for area_dict in permit_areas_dict:
             permit_area, _ = PermitArea.objects.update_or_create(
                 domain=EnforcementDomain.get_default_domain(),
                 identifier=area_dict['identifier'],
+                permitted_user=get_user_model().objects.get_or_create(
```
suutari-ai: The area importer shouldn't create the user IMHO, so this should be just `.filter(username=...).get()`.

urshala:
LGTM 👍

PR #119, opened 2020-01-30
Multicity endpoints

urshala:
* Refactoring of code so that the common endpoints could be shared between enforcement and operator endpoints
* Add current operator as owner of permitseries while creating permitseries
* Add endpoint to list enforcement domain to Operator API
* Add endpoint to CRUD permit to Operator API
* Add endpoint to activate/deactivate permitseries to Operator API
* Add endpoint to CRUD permitseries to Operator API
* Add domain field to parkings.

### DIFF
--- a/parkings/api/common_permit.py
+++ b/parkings/api/common_permit.py
@@ -101,6 +101,9 @@ class PermitViewSet(viewsets.ModelViewSet):
             kwargs['many'] = True
         return super().get_serializer(*args, **kwargs)
 
+    def get_queryset(self):
+        return super().get_queryset().filter(series__owner=self.request.user)
+
 
 class ActivePermitByExternalIdSerializer(PermitSerializer):
     class Meta(PermitSerializer.Meta):
--- a/parkings/api/common_permit.py
+++ b/parkings/api/common_permit.py
@@ -29,6 +29,9 @@ class PermitSeriesViewSet(CreateAndReadOnlyModelViewSet):
     queryset = PermitSeries.objects.all()
     serializer_class = PermitSeriesSerializer
 
+    def perform_create(self, serializer):
+        serializer.save(owner=self.request.user)
+
     @action(detail=True, methods=['post'])
     def activate(self, request, pk=None):
         with transaction.atomic():
--- a/parkings/api/common_permit.py
+++ b/parkings/api/common_permit.py
@@ -55,6 +55,9 @@ class PermitSeriesViewSet(CreateAndReadOnlyModelViewSet):
 
             return Response({'status': 'OK'})
 
+    def get_queryset(self):
+        return super().get_queryset().filter(owner=self.request.user)
+
 
 class PermitListSerializer(serializers.ListSerializer):
     def create(self, validated_data):
--- a/parkings/api/common_permit.py
+++ b/parkings/api/common_permit.py
@@ -1,0 +1,117 @@
+from django.db import transaction
+from django.utils.translation import gettext as _
+from django_filters.rest_framework import DjangoFilterBackend
+from rest_framework import mixins, permissions, serializers, viewsets
+from rest_framework.decorators import action
+from rest_framework.exceptions import NotFound
+from rest_framework.response import Response
+
+from ..models import EnforcementDomain, Permit, PermitSeries
+
+
+class PermitSeriesSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = PermitSeries
+        fields = ['id', 'created_at', 'modified_at', 'active']
+        read_only_fields = fields
+
+
+class CreateAndReadOnlyModelViewSet(
+        mixins.CreateModelMixin,
+        mixins.RetrieveModelMixin,
+        mixins.ListModelMixin,
+        viewsets.GenericViewSet):
+    pass
+
+
+class PermitSeriesViewSet(CreateAndReadOnlyModelViewSet):
+    permission_classes = [permissions.IsAdminUser]
+    queryset = PermitSeries.objects.all()
+    serializer_class = PermitSeriesSerializer
+
+    @action(detail=True, methods=['post'])
+    def activate(self, request, pk=None):
+        with transaction.atomic():
+            obj_to_activate = self.get_object()
+            old_actives = self.queryset.filter(active=True)
+
+            if obj_to_activate.active and old_actives.count() == 1:
+                return Response({'status': 'No change'})
+
+            if not obj_to_activate.active:
+                obj_to_activate.active = True
+                obj_to_activate.save()
+
+            for obj in old_actives.exclude(pk=obj_to_activate.pk):
+                obj.active = False
+                obj.save()
+
+            prunable_series = PermitSeries.objects.prunable()
+            Permit.objects.filter(series__in=prunable_series).delete()
+            prunable_series.delete()
+
+            return Response({'status': 'OK'})
+
+
+class PermitListSerializer(serializers.ListSerializer):
+    def create(self, validated_data):
+        permits = [Permit(**item) for item in validated_data]
+        return Permit.objects.bulk_create(permits)
+
+
+class PermitSerializer(serializers.ModelSerializer):
+    class Meta:
+        list_serializer_class = PermitListSerializer
+        model = Permit
+        fields = [
+            'id',
+            'series',
+            'external_id',
+            'subjects',
+            'areas',
+        ]
+        read_only_fields = ['id']
+
+    def validate(self, attrs):
+        if self.instance is None:
+            if attrs.get('domain', None) is None:
+                attrs['domain'] = EnforcementDomain.get_default_domain()
+            instance = Permit(self.instance, **attrs)
+            instance.clean()
+        else:
+            instance = self.instance
+            for key in attrs:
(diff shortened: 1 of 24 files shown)


parkings/api/common_permit.py
```diff
@@ -15,6 +15,10 @@ class Meta:
         fields = ['id', 'created_at', 'modified_at', 'active']
         read_only_fields = fields
 
+    def validate(self, data):
+        data['owner'] = self.context['request'].user
```
suutari-ai: I think there should be some better place to do this, since the owner field should only be touched on creation, not on all validation. How about doing this in the `create` method?

parkings/api/operator/operator_permission.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework import permissions
```
suutari-ai: Not related to this line, but since GitHub doesn't allow commenting the commit messages putting the comment here:
> Refactor OperatorApi permission
>
> Refactor OpertorAPI has_permission as this will be shared among
> other Operator endpoints too.

This commit is just moving the code, so the title would be more descriptive as "Move OperatorApiHasPermission to separate module" (or something similar). Just "Refactor something" doesn't really tell what kind of a refactoring it is (could be e.g. move, rename, split to pieces, combine from pieces, complete rewrite, reordering methods, etc.).

suutari-ai: Sorry, misread the code first. It wasn't just moving the class, but actually splitting out a base class from it. So a better title would be e.g. "Operator API: Split out common base class for permissions" or similar.

parkings/api/operator/urls.py
```diff
@@ -1,9 +1,25 @@
...
+from .enforcement_domain import EnforcementDomainView
 from .parking import OperatorAPIParkingViewSet
 
-router = DefaultRouter()
+
+class Router(DefaultRouter):
+    def get_urls(self):
+        urls = super().get_urls()
+        return urls + [
+            url(r"^enforcement_domain/$", EnforcementDomainView.as_view(), name="enforcement_domain")
```
suutari-ai: Use a ViewSet for this rather than a custom view. That would allow using a normal route.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,38 @@
...
+    class Meta(PermitSerializer.Meta):
+        fields = PermitSerializer.Meta.fields + ['domain']
+
+
+class FilterUserOwnedDomainPermitMixin():
+    def get_queryset(self):
+        return (
+            super()
+            .get_queryset()
+            .filter(domain__enforcer__user=self.request.user)
```
suutari-ai: Operators should see all their permits. It has nothing to do with enforcers.

You should filter based on the PermitSeries.owner field:
```python
qs = super().get_queryset()
return qs.filter(series__owner=self.request.user)
```

suutari-ai: Since the tests are passing. This also means that there is no good test coverage for cases like this, so they should probably be updated and some new tests added too.

suutari-ai: E.g. here's a couple test cases for this.

Test case 1:

1. Operator A creates a PermitSeries PS1 and permits P1 and P2 into it.
2. Operator B creates a PermitSeries PS2 and permits P3 and P4 into it.
3. Operator A fetches list of all permits. It should see exactly P1 and P2.
4. Operator B fetches list of all permits. It should see exactly P3 and P4.

Test case 2:

Setup: There is an operator A, an enforcement domains X and Y, and an enforcer E which is enforcing the X.
1. Operator A creates a PermitSeries PS1 and permits P1 and P2 into it, with P1.domain = X, P2.domain = Y and they are both for registration_number ABC-123.
2. Enforcer E creates a PermitSeries PS2 and permits P3 and P4 into it. *Note*: Happens through Enforcement API, so the domains are set automatically as P3.domain = P4.domain = X. Registration number for these too is ABC-123.
3. Operator A fetches list of all permits. It should see exactly P1 and P2.
4. Enforcer E fetches list of all permits through the permit management API (/enforcement/v1/permit/). It should see exactly P3 and P4. (The operator Permits are not listed here, because this endpoint is not meant for checking the permits, but for managing the enforcers own permits.)
5. (*Later* - when the checking endpoint is updated) Enforcer E fetches the list of valid parkings for ABC-123. The result should include data from permits P1, P3 and P4, but *not* P2, since it's for another domain.

parkings/tests/api/conftest.py
```diff
@@ -57,3 +57,13 @@ def operator_2_api_client(operator_2):
...
+
+
+@pytest.fixture
+def operator_enforcer(enforcer_factory, operator):
+    return enforcer_factory(user=operator.user)
+
+
+@pytest.fixture
+def operator_enforcer2(enforcer_factory, operator_2):
+    return enforcer_factory(user=operator_2.user)
```
suutari-ai: These two shouldn't be needed as there should be no operators that are also enforcers.

urshala:
@suutari-ai I have made the initial suggested changes. Please have a look it.

parkings/api/operator/permit.py
urshala: The second last test, number 4 is not included in this PR. I think it should belong to separate commit as we will be making changes to `EnforcementPermitViewSet`.

parkings/tests/api/operator/test_permit.py
```diff
@@ -0,0 +1,270 @@
...
+    assert obj.id == data['id']
+    assert obj.subjects == data['subjects']
+    assert obj.areas == data['areas']
+
+
+def test_operator_can_create_permit_with_valid_post_data(
+    operator_api_client, permit_series, enforcement_domain
+):
+    permit_data = _get_permit_data(permit_series)
+    permit_data.update(domain=enforcement_domain.id)
```
suutari-ai: The endpoint should take the **code** of the Enforcement Domain, not its id in the `domain` field. And since all operator permit endpoints will need the domain in the data, you could as well hard code it in `_get_permit_data` as `'domain': 'ED1',` or something similar (and make sure that the enforcement domain code is ED1).

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,29 @@
...
+
+class DomainToCodeMixin():
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation.update(domain=instance.domain.code)
+        return representation
+
+
+class OperatorPermitSerializer(
+    DomainNotRequiredMixin,
```
suutari-ai: Domain should be required in the Operator API.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,29 @@
...
+class DomainNotRequiredMixin():
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['domain'].required = False
+
+
+class DomainToCodeMixin():
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation.update(domain=instance.domain.code)
```
suutari-ai: Use `SlugRelatedField` for the `domain` field so you don't have to do this conversion manually and it will handle the other way also. https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield

suutari-ai:
```python
class OperatorPermitSerializer(...):
    domain = serializers.SlugRelatedField(
        slug_field='code', queryset=EnforcementDomain.objects.all())
```

parkings/tests/api/operator/test_permit.py
```diff
@@ -0,0 +1,259 @@
...
+    operator_enforcer = enforcer_factory(user=operator.user)
+    permit_data = _get_permit_data(permit_series)
+    permit_data.update(domain=operator_enforcer.enforced_domain.code)
+
+    PermitArea.objects.create(
+        name='Kamppi',
+        identifier='A',
+        geom=area,
+        permitted_user=operator.user
+    )  # PermitArea with different domain compared to permit_data
```
suutari-ai: How can this PermitArea be created here without the domain?

parkings/tests/api/operator/test_permit.py
```diff
@@ -0,0 +1,259 @@
...
+
+    for each_area in permit_data['areas']:
+        each_area.update(area='A')
+
+    response = operator_api_client.post(list_url, data=permit_data)
+
+    assert response.json()['non_field_errors'] == [
+        'You can\'t create permit whose domain is different than areas\'s domain.'
```
suutari-ai: Hmm.. the permit area is looked up by the (domain, area identifier) pair, so it's not possible by definition to create a permit which references an area in a different domain. Either the pair (HKI, A) is a valid key for a permit area or it's not.

parkings/tests/api/operator/test_permit.py
urshala: The domain is populated while saving PermitArea in model.

suutari-ai: Hmm.. OK. domain shouldn't have a default value for those objects. The default for domain field is needed only for Parking objects. Gotta check that too.

Anyway, here, the domain should be specified.

suutari-ai:
Replaced by #123, #124, #125, and #126.

PR #120, opened 2020-02-10
Valid parkings with permit

amanpdyadav:
And endpoint to return the list of valid parking along with the permit.


parkings/api/enforcement/valid_parking.py
```diff
@@ -141,3 +149,26 @@ def get_grace_duration(default=datetime.timedelta(minutes=15)):
     value = getattr(settings, 'PARKKIHUBI_TIME_OLD_PARKINGS_VISIBLE', None)
     assert value is None or isinstance(value, datetime.timedelta)
     return value if value is not None else default
+
+
+class ValidParkingPermitViewSet(ValidParkingViewSet):
```
kerkkoheiskanen, line 154: Unifying the naming here might be good, see `ValidParkingAndPermitSerializer `, which suggests that the name for this class should be `ValidParkingAndPermitViewSet `.

kerkkoheiskanen:
There should be a test testing the functionality of ValidParkingPermitViewSet, as it stands the PR adds an endpoint but doesn't add a dedicated test for it.

parkings/api/enforcement/valid_parking.py
```diff
@@ -41,6 +42,13 @@ def to_representation(self, instance):
         return representation
 
 
+class ValidParkingAndPermitSerializer(ValidParkingSerializer, serializers.Serializer):
```
kerkkoheiskanen, line 45: Perhaps the new endpoint changes should be in a different commit, as they are unrelated to the are field changes.

suutari-ai:
* There is already check_parking endpoint for querying both parkings and permits: https://api.parkkiopas.fi/docs/enforcement/#operation/checkParking
* Permits can be queried with valid_permit_item endpoint: https://api.parkkiopas.fi/docs/enforcement/#operation/getValidPermitItemList

PR #124, opened 2020-02-14
Parking visibility by domain

urshala:
* Add domain field to parkings
* Allow enforcers to view only parkings of its enforced domain

### DIFF
--- a/parkings/api/enforcement/valid_parking.py
+++ b/parkings/api/enforcement/valid_parking.py
@@ -4,9 +4,10 @@ import django_filters
 from django.conf import settings
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
-from rest_framework import permissions, serializers, viewsets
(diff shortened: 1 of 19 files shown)


parkings/api/enforcement/valid_parking.py
```diff
@@ -136,6 +136,9 @@ def _get_filterset(self, queryset):
         filter_backend = self.filter_backends[0]()
         return filter_backend.get_filterset(self.request, queryset, self)
 
+    def get_queryset(self):
+        return super().get_queryset().filter(domain=self.request.user.enforcer.enforced_domain)
```
suutari-ai: Add a separate commit before this commit for `IsEnforcer` permission class and putting it into use for all Enforcement API endpoints, since otherwise we cannot be sure that  `request.user.enforcer` exists.

suutari-ai:
Looks good now. :+1: Thanks, Deepak!

PR #126, opened 2020-02-16
Operator permit endpoints

urshala:
* Refactor Enforcement API permit endpoints to be reusable in Operator API
* Split the IsOperator permission to make it reusable
* Add permitseries endpoint 
* Add endpoint to activate permitseries
* Add permit endpoint
* Add activepermit endpoint

### DIFF
--- a/parkings/api/operator/permit.py
+++ b/parkings/api/operator/permit.py
@@ -3,9 +3,10 @@ from rest_framework import mixins, serializers
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
-from parkings.models import Permit, PermitSeries
+from parkings.models import EnforcementDomain, Permit, PermitSeries
 
-from ..common_permit import PermitSeriesViewSet
+from ..common_permit import (
+    PermitSerializer, PermitSeriesViewSet, PermitViewSet)
 from .permissions import IsOperator
 
 
@@ -50,3 +51,16 @@ class PermitSeriesActivateBodySerializer(serializers.Serializer):
     deactivate_series = serializers.ListField(
         child=serializers.IntegerField(), required=False
     )
+
+
+class OperatorPermitSerializer(PermitSerializer):
+    domain = serializers.SlugRelatedField(
+        slug_field='code', queryset=EnforcementDomain.objects.all())
+
+    class Meta(PermitSerializer.Meta):
+        fields = PermitSerializer.Meta.fields + ['domain']
+
+
+class OperatorPermitViewSet(PermitViewSet):
+    serializer_class = OperatorPermitSerializer
+    permission_classes = [IsOperator]
--- a/parkings/api/operator/permit.py
+++ b/parkings/api/operator/permit.py
@@ -1,4 +1,9 @@
-from rest_framework import mixins
+from django.db import transaction
+from rest_framework import mixins, serializers
+from rest_framework.decorators import action
+from rest_framework.response import Response
+
+from parkings.models import Permit, PermitSeries
 
 from ..common_permit import PermitSeriesViewSet
 from .permissions import IsOperator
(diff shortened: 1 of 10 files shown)


parkings/api/operator/permit.py
```diff
@@ -1,8 +1,64 @@
...
+        if 'deactivate_series' in data:
+            data['deactivate_others'] = False
+
+        if (
+            'deactivate_others' in data
+            or 'deactivate_series' in data
+        ):
+            return super().validate(data)
+
+        raise serializers.ValidationError(
```
urshala: TODO: The default value should be {} i.e. Not to deactivate any other series.

parkings/api/operator/permit.py
```diff
@@ -1,8 +1,64 @@
...
+            old_actives = self.queryset.filter(active=True, owner=self.request.user)
+
+            # Always activate the specified permit series regardless of what to deactivate
+            if not obj_to_activate.active:
+                obj_to_activate.active = True
+                obj_to_activate.save()
+
+            deactivate_payload = OperatorPermitSeriesPayload(data=request.data)
+            deactivate_payload.is_valid(raise_exception=True)
+            validated_data = deactivate_payload.validated_data
```
suutari-ai: This should be done first, even before starting the transaction.

parkings/api/operator/permit.py
```diff
@@ -1,8 +1,64 @@
...
 
 
 class OperatorPermitSeriesViewSet(mixins.DestroyModelMixin, PermitSeriesViewSet):
     permission_classes = [IsOperator]
+
+    @action(detail=True, methods=['post'])
+    def activate(self, request, pk=None):
+        with transaction.atomic():
+            obj_to_activate = self.get_object()
+            old_actives = self.queryset.filter(active=True, owner=self.request.user)
```
suutari-ai: I think `self.get_queryset()` should be used here and the owner filtering should happen there.

parkings/api/operator/permit.py
suutari-ai: Yep, empty body should be allowed too.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+
+            prunable_series = PermitSeries.objects.prunable()
+            Permit.objects.filter(series__in=prunable_series).delete()
+            prunable_series.delete()
+
+            return Response({'status': 'OK'})
+
+
+class OperatorPermitSeriesPayload(serializers.Serializer):
+    deactivate_others = serializers.BooleanField(required=False)
```
suutari-ai: Use `default=False` here.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+    deactivate_series = serializers.ListField(
+        child=serializers.IntegerField(), required=False
+    )
+
+    def validate(self, data):
```
suutari-ai: I think you don't have to specify validate method at all, if you just set the default values correctly in the field serializers.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+                ).update(active=False)
+
+            prunable_series = PermitSeries.objects.prunable()
+            Permit.objects.filter(series__in=prunable_series).delete()
+            prunable_series.delete()
+
+            return Response({'status': 'OK'})
+
+
+class OperatorPermitSeriesPayload(serializers.Serializer):
```
suutari-ai: I think better name for this could be e.g. `PermitSeriesActivateBodySerializer` since it's a serializer of the request body of the permitseries/activate endpoint.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+            old_actives = self.queryset.filter(active=True, owner=self.request.user)
+
+            # Always activate the specified permit series regardless of what to deactivate
+            if not obj_to_activate.active:
+                obj_to_activate.active = True
+                obj_to_activate.save()
+
+            deactivate_payload = OperatorPermitSeriesPayload(data=request.data)
+            deactivate_payload.is_valid(raise_exception=True)
+            validated_data = deactivate_payload.validated_data
```
suutari-ai: This body data validation should be done first, even before starting the transaction.

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+            _('Either deactivate_others or deactivate_series is required')
+        )
+
+
+class OperatorPermitSerializer(PermitSerializer):
+    domain = serializers.SlugRelatedField(
+        slug_field='code', queryset=EnforcementDomain.objects.all())
+
+    class Meta(PermitSerializer.Meta):
+        fields = PermitSerializer.Meta.fields + ['domain']
```
suutari-ai: What happens if you leave out this Meta and fields? (Just asking, because I thought that this would be unneeded, since the field is explicitly defined above.)

parkings/api/operator/permit.py
```diff
@@ -0,0 +1,92 @@
...
+    serializer_class = OperatorPermitSerializer
+    permission_classes = [IsOperator]
+
+
+class OperatorActivePermitByExtIdSerializer(ActivePermitByExternalIdSerializer):
+    domain = serializers.SlugRelatedField(
+        slug_field='code', queryset=EnforcementDomain.objects.all())
+
+    class Meta(ActivePermitByExternalIdSerializer.Meta):
+        fields = ActivePermitByExternalIdSerializer.Meta.fields + ['domain']
```
suutari-ai: [Same here](https://github.com/City-of-Helsinki/parkkihubi/pull/126/files#r381181520): What happens if you leave out this Meta and fields?

parkings/tests/api/operator/test_permit_series.py
```diff
@@ -0,0 +1,188 @@
...
+    url = reverse('operator:v1:permitseries-detail', kwargs={'pk': permit_series.pk})
+    response = operator_api_client.delete(url)
+
+    assert response.status_code == HTTP_204_NO_CONTENT
+    assert PermitSeries.objects.count() == 1
+    assert PermitSeries.objects.first().id != permit_series.pk
+
+
+def get_activate_url(obj):
+    return reverse('operator:v1:permitseries-detail', kwargs={'pk': obj.pk})
```
suutari-ai: This is getting the object url, not activate url. Either append the `activate/` here, or rename the function.

parkings/api/operator/permit.py
urshala: It throws an error saying that `domain` is declared but not added to fields.

suutari-ai:
Thanks Deepak!

PR #130, opened 2020-02-26
Add parking forecast model training and predictions

klemmari1:
- Add some dependencies
- Functions for training a gradient boosting model to forecast parking counts based on historical data and predicting parking counts 1 week into the future using the trained model.
- Gradient boosting forecast model implemented with CatBoost library.
- Add cron jobs that caches the parking counts and makes the predictions daily. By default the jobs are turned off.

### DIFF
--- a/parkings/cron.py
+++ b/parkings/cron.py
@@ -1,0 +1,29 @@
+from django.conf import settings
+from django_cron import CronJobBase, Schedule
+
+from parkings.forecasts.cache_parking_counts import (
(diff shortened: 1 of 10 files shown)


suutari-ai:
I don't know why this was left hanging here.

Since this is not even going to `master` branch, it's better to just merge this even if there could be some open questions how to utilize this feature in final product. Those open questions can/should be solved when merging from the `forecast` branch to `master`.

PR #133, opened 2020-04-01
Make parkings migrations run faster

japauliina:
The parkings migration 0030 took way too long to run (hours). Optimise the queries to only add domains to valid parkings. Now when testing with about 4000 valid parkings the migration took
less than a minute.

For this change to work enable `null=True` for domains in Parking model.


suutari-ai:
Some info about how long it takes now would be nice (in the commit message).

parkings/migrations/0031_default_domain.py
```diff
@@ -20,7 +20,9 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='parking',
             name='domain',
-            field=models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='parkings', to='parkings.EnforcementDomain'),
+            field=models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='parkings',
+                to='parkings.EnforcementDomain', blank=True, null=True,
+            ),
```
suutari-ai, line 25: Is this AlterField still needed? What it changes (compared to previous version of the same field)?

suutari-ai: The previous version is here:

https://github.com/City-of-Helsinki/parkkihubi/blob/e19d6777a7c08edae4bb94107807a08a35c83bd0/parkings/migrations/0029_enforcement_domain.py#L75-L79

It seems that only difference is blank=True, but that can as well be left out. So, this whole AlterField command can be removed from this migration.

parkings/models/parking.py
```diff
@@ -80,7 +80,7 @@ class Parking(TimestampedModelMixin, UUIDPrimaryKeyMixin):
     )
     domain = models.ForeignKey(
         EnforcementDomain, on_delete=models.PROTECT,
-        related_name='parkings')
+        related_name='parkings', null=True, blank=True,)
```
suutari-ai, line 83: blank=True is not needed. Or if you leave it here, then add it to 0029 too to avoid changing the field twice in migrations.

PR #134, opened 2020-04-01
PermitLookUpItem: Change area relation from identifier to ForeignKey

kerkkoheiskanen:
This change was originally commited in 478b0f6 but after changes
to multicity branch there were conflicts which are resolved in this
Merge Request. In addition, parking and permit api changes were omitted
from this commit to make the commits more atomic.


parkings/models/permit.py
```diff
@@ -157,7 +157,7 @@ def _make_lookup_items(self):
                     permit=self,
                     registration_number=Parking.normalize_reg_num(
                         subject['registration_number']),
-                    area_identifier=area['area'],
+                    area=PermitArea.objects.get(identifier=area['area']),
```
suutari-ai, line 160: Shouldn't this use the domain as part of the lookup key too?

parkings/models/permit.py
```diff
@@ -175,14 +175,14 @@ def by_subject(self, registration_number):
         return self.filter(registration_number=normalized_reg_num)
 
     def by_area(self, area_identifier):
-        return self.filter(area_identifier=area_identifier)
+        return self.filter(area__identifier=area_identifier)
```
suutari-ai, line 178: Here too, should either take the domain as an argument like `def by_area(domain, area_identifier)` or an area object `def by_area(area): assert isinstance(area, PermitArea)`.

PR #139, opened 2020-04-06
Fix and clean-up Permit related code

suutari-ai:
There was some bugs and unhandled edge cases in the code related to
Permits, PermitAreas and PermitSeries.  Fix the bugs, improve the
factories and clean-up the code a bit.

### DIFF
--- a/parkings/tests/api/enforcement/test_check_parking.py
+++ b/parkings/tests/api/enforcement/test_check_parking.py
@@ -31,6 +31,23 @@ INVALID_PARKING_DATA = {
 }
 
 
+def create_payment_zone(client=None, domain=None):
+    assert client or domain
+    return PaymentZone.objects.create(
+        domain=(domain or client.enforcer.enforced_domain), code="1",
+        number=1, name="Maksuvyöhyke 1", geom=create_area_geom())
+
+
+def create_permit_area(client=None, domain=None, permitted_user=None):
+    assert client or (domain and permitted_user)
+    return PermitArea.objects.create(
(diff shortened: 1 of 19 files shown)


parkings/tests/api/enforcement/test_check_parking.py
```diff
@@ -30,6 +31,23 @@
...
 
 
+def create_payment_zone(client=None, domain=None):
+    assert client or domain
+    return PaymentZone.objects.create(
+        domain=(domain or client.enforcer.enforced_domain), code="1",
+        number=1, name="Maksuvyöhyke 1", geom=create_area_geom())
+
+
+def create_permit_area(client=None, domain=None, permitted_user=None):
```
kerkkoheiskanen, line 41: Could `factories.permit.create_permit_area()` be used here? Now there's two functions with the same name that do the same thing. Client can be asserted before `create_permit_area()` is called here and you can just give `permitted_user or client.auth_user` to `create_permit_area()` as an argument.

suutari-ai: I think it's important here in these tests that the PermitArea has this particular geometry. That's why I didn't use the factories.permit.create_permit_area.

suutari-ai: And also the identifier must be "A" for these tests. I think it's justified to use a local version of create_permit_area, since the general one doesn't fit for this test file's use cases.

kerkkoheiskanen: Ah I didn't notice the geometry. I guess it's justified then.

PR #141, opened 2020-04-07
WIP: Enforcement: Enable fetching of all valid parkings

kerkkoheiskanen:
Remove registration number as a required parameter,
enabling all valid parkings to be fetched from valid parking
endpoint.


japauliina:
Finished in #145 & #149.

PR #144, opened 2020-04-08
Operator api: Add endpoint for PaymentZones

japauliina:
Add a new endpoint for listing PaymentZones for permitted operators.
This change has already been documented in fd7bf05.

Also, fix an `UnorderedObjectListWarning` by adding an ordering for the PaymentZone model.

### DIFF
--- a/parkings/models/zone.py
+++ b/parkings/models/zone.py
@@ -19,6 +19,7 @@ class PaymentZone(TimestampedModelMixin, UUIDPrimaryKeyMixin):
 
     class Meta:
         unique_together = [('domain', 'code')]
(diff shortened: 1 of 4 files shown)


parkings/models/zone.py
```diff
@@ -19,6 +19,7 @@ class PaymentZone(TimestampedModelMixin, UUIDPrimaryKeyMixin):
 
     class Meta:
         unique_together = [('domain', 'code')]
+        ordering = ('number',)
```
suutari-ai, line 22: `('domain', 'code')` might be better ordering for these.

PR #145, opened 2020-04-15
Enforcement: Add documentation for new endpoints

kerkkoheiskanen:
Add documentation for new endpoints for fetching valid permits
and valid parkings.

### DIFF
--- a/docs/api/enforcement.yaml
+++ b/docs/api/enforcement.yaml
@@ -146,6 +146,23 @@ components:
           $ref: '#/components/schemas/PermitSubjects'
         areas: &permitAreas
           $ref: '#/components/schemas/PermitAreas'
+    ValidPermitItem:
+      type: object
+      properties:
+        area:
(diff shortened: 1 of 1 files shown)


docs/api/enforcement.yaml
```diff
@@ -146,6 +146,34 @@ components:
...
+      properties:
+        id:
+          description: Permit id
+          type: integer
+        created_at:
+          description: Permit creation date
+          type: string
+          format: date-time
+        modified_at:
+          description: Permit modified date
```
suutari-ai, line 160: nit: "modification" to be in line with the "creation date" above

docs/api/enforcement.yaml
```diff
@@ -146,6 +146,34 @@ components:
...
+          type: string
+          format: date-time
+        Area:
```
suutari-ai, line 163: should be lower case, i.e. `area`

PR #146, opened 2020-04-15
Parking: Change payment zone to foreignkey

kerkkoheiskanen:
In this PR, Parking model's field zone is changed from `IntegerField` to PaymentZone `ForeignKey`.

In the Parking endpoints, the zone field should return related PaymentZone code. However, since this field returns a string, for backward compatibility also integer should be accepted. The parking endpoints should return code as integer wherever possible.


suutari-ai:
Squash these two commits together, since the first one breaks tests without the second one and it's not OK to break the tests in any commit.

> Parking: Change zone field from int to foreignkey

> Parking: Fix parts broken by zone field change

parkings/api/enforcement/valid_parking.py
```diff
@@ -29,9 +30,19 @@ class Meta:
             'is_disc_parking',
         ]
 
+    def __init__(self, *args, **kwargs):
+        super(ValidParkingSerializer, self).__init__(*args, **kwargs)
+        domain = self.context['request'].user.enforcer.enforced_domain
+        zone_field = self.fields['zone']
+        zone_qs = zone_field.get_queryset()
+        if zone_qs:
+            zone_qs.queryset = zone_qs.filter(domain=domain)
```
suutari-ai, line 39: Shouldn't this be `zone_field.queryset = ...`?

I'm pretty sure that it has a bug. Maybe add a test to prove it:
1. Add two zones: one to Domain D1 and another to Domain D2, both with the same code "Z"
2. Create a parking P1 to D1 with zone code Z
3. Check the zone FK of P1
4. Create a parking P2 to D2 with zone code Z
5. Check the zone FK of P2

suutari-ai: Oh, it was actually the valid_parking endpoint, not the parking creating endpoint (which I was talking about here). Well, then the test case should be almost the same, but instead of checking the created object, you should be checking that the valid_parking returns correct results when two parkings exists with the same zone code, but in different domains.

japauliina: Yes, it should be `zone_field.queryset` and not `zone_qs.queryset`.. Will fix this and add the tests.

japauliina: Bug fixed and tests added for both endpoints.

parkings/models/zone.py
```diff
@@ -23,3 +23,9 @@ class Meta:
 
     def __str__(self):
         return self.name
+
+    def get_code(self):
+        try:
+            return int(self.code)
+        except ValueError:
+            return self.code
```
suutari-ai, line 31: This function could probably have a more descriptive name, and maybe it should be a property, since it's really light computation, not really "getting" or "fetching" anything from DB or anywhere. Also, I would use isdigit to avoid the try-catch. How about something like this?
```python
@property
def casted_code(self):
    return int(self.code) if self.code.isdigit() else self.code
```

parkings/api/monitoring/serializers.py
```diff
@@ -25,3 +26,8 @@ class Meta:
             'modified_at',
         ]
         geo_field = 'location'  # Stored already in SRID 4326 / WGS84
+
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation['properties']['zone'] = instance.zone.get_code()
```
suutari-ai, line 32: This is not handling the case of zone being None. (disc parkings)

parkings/tests/api/operator/test_disc_parking.py
```diff
@@ -34,6 +35,7 @@ def disc_parking_data():
 
 
 def test_post_disc_parking(operator_api_client, operator, disc_parking_data):
+    create_enforced_domain_for_user(operator.user)
```
suutari-ai, line 38: I don't understand why operator would have to be an enforcer here. Are you sure this enforcer-creation is needed?

parkings/tests/api/operator/test_disc_parking.py
```diff
@@ -45,6 +47,7 @@ def test_post_disc_parking(operator_api_client, operator, disc_parking_data):
 
 
 def test_end_disc_parking_with_patch(operator_api_client, operator, disc_parking):
+    create_enforced_domain_for_user(operator.user)
```
suutari-ai, line 50: Same here. Is this needed? https://github.com/City-of-Helsinki/parkkihubi/pull/146/commits/d2cd37559501da7fbec0130bfe7d4f9ad860b437#r412981973

parkings/tests/api/operator/test_disc_parking.py
```diff
@@ -64,5 +67,6 @@ def test_delete_disc_parking(operator_api_client, disc_parking):
 
 
 def test_required_fields_for_disc_parking(operator_api_client, operator, disc_parking_data):
+    create_enforced_domain_for_user(operator.user)
```
suutari-ai, line 70: Same here. Is this needed? https://github.com/City-of-Helsinki/parkkihubi/pull/146/commits/d2cd37559501da7fbec0130bfe7d4f9ad860b437#r412981973

parkings/tests/api/operator/test_parking.py
```diff
@@ -104,7 +104,8 @@ def test_unauthenticated_and_normal_users_cannot_do_anything(api_client, user_ap
     check_method_status_codes(user_api_client, urls, ALL_METHODS, 403, error_code='permission_denied')
 
 
-def test_parking_required_fields(operator_api_client, parking):
+def test_parking_required_fields(operator_api_client, operator, parking):
+    create_enforced_domain_for_user(operator.user)
```
suutari-ai, line 108: Same here. Is this needed? https://github.com/City-of-Helsinki/parkkihubi/pull/146/commits/d2cd37559501da7fbec0130bfe7d4f9ad860b437#r412981973

parkings/tests/api/operator/test_parking.py
```diff
@@ -390,3 +399,41 @@ def test_enforcement_domain_can_be_specified_on_parking_creation(operator_api_cl
...
+    create_payment_zone(code=str(new_parking_data['zone']), domain=domain)
+    response_parking_data = post(operator_api_client, list_url, new_parking_data)
+    assert response_parking_data['zone'] == '5A'
+
+
+def test_zone_in_correct_domain_is_set_for_parking(
+        operator_api_client, operator, operator_2_api_client, operator_2, new_parking_data
+):
+    domain_1 = create_enforced_domain_for_user(operator.user)
+    domain_2 = create_enforced_domain_for_user(operator_2.user)
```
suutari-ai, line 427: These domain's shouldn't be enforced by the operators.

Operators create the parkings and Enforcers check them. Operators are allowed to create parkings to whichever enforcement domain, but enforcers are allowed to see only the parkings in their own domain.

parkings/api/enforcement/valid_parking.py
```diff
@@ -29,9 +30,19 @@ class Meta:
             'is_disc_parking',
         ]
 
+    def __init__(self, *args, **kwargs):
+        super(ValidParkingSerializer, self).__init__(*args, **kwargs)
+        domain = self.context['request'].user.enforcer.enforced_domain
```
suutari-ai, line 35: Use the domain of the parking here, no the domain of the user executing the request. Although in valid_parking endpoint it doesn't matter so much, since they should be the same (since the enforcer executing the request should have access only to those parkings which are in their domain).

parkings/api/operator/parking.py
```diff
@@ -49,6 +49,12 @@ def __init__(self, *args, **kwargs):
 
         self._set_required_extra_fields()
 
+        domain = self.context['request'].user.enforcer.enforced_domain
```
suutari-ai, line 52: The user is not supposed to be an Enforcer here. It's an operator. Use the domain of the parking as the domain on this line. Though it probably can't be done like this, since the parking is not yet created/initialized.

parkings/tests/api/operator/test_parking.py
japauliina: Seems like I had a little misunderstanding about the operator and domain link, this clears it! I’ll check the related code again and fix these 🙂

parkings/api/enforcement/valid_parking.py
japauliina: I think this was not needed after all since the queryset filter already covered the domain check.

parkings/api/operator/parking.py
japauliina: How can we currently know when posting the new parking what the domain should be?

parkings/api/operator/parking.py
```diff
@@ -101,18 +104,22 @@ class OperatorAPIParkingViewSet(mixins.CreateModelMixin, mixins.UpdateModelMixin
     serializer_class = OperatorAPIParkingSerializer
 
     def _get_zone(self):
+        if self.request.data.get('domain'):
+            domain = EnforcementDomain.objects.get(code=self.request.data.get('domain'))
+        else:
+            domain = EnforcementDomain.get_default_domain()
```
suutari-ai, line 110: We can avoid a database query here by not looking up the EnforcementDomain object from the database. Let's rather just take the domain_code to a variable as a string and then use that in the query below, like: `PaymentZone.objects.get(domain__code=domain_code, code=zone_code)`.

But that optimization change can be done in another PR to get this merged already.

parkings/api/operator/parking.py
suutari-ai: Already replied in Slack, but repeating here for future reference. It's supposed to be part of the parameters in the request, or if it's left out, the default value is from settings (as used by EnforcementDomain.get_default_domain function).

parkings/api/enforcement/valid_parking.py
suutari-ai: I'm a bit uncertain if the slug field will work without the filtering if there are two zones with the same code, but we can add a test case for this in another PR.

parkings/api/operator/parking.py
suutari-ai: I ended up rewriting this with SlugRelatedField, since it seemed to make the code cleaner: https://github.com/City-of-Helsinki/parkkihubi/pull/151

parkings/api/enforcement/valid_parking.py
suutari-ai: Ah, now I see why it works. However I think it might be better to go with a more explicit version, see #150.

PR #153, opened 2020-04-27
Monitoring: Limit parking visibility by domain

japauliina:
Users of the monitoring API are bound to specific Enforcement Domain and the parkings they see is limited to that particular domain.


parkings/api/monitoring/valid_parking.py
```diff
@@ -38,3 +38,6 @@ class ValidParkingViewSet(viewsets.ReadOnlyModelViewSet):
     bbox_filter_field = 'location'
     filter_backends = [DjangoFilterBackend, WGS84InBBoxFilter]
     bbox_filter_include_overlapping = True
+
+    def get_queryset(self):
+        return super().get_queryset().filter(domain=self.request.user.monitor.domain)
```
suutari-ai, line 43: You need to make sure that self.user has a Monitor object linked to it. This is done in the Enforcement and Operator API endpoints in the permission class that is linked to the viewset via the `permission_classes` keyword. So check if you can add similar code to the `MonitoringApiPermission` which is used here.

PR #163, opened 2020-05-06
Improve multicity migrations 0032 and 0035 (HP-109)

suutari-ai:
In migration 0032 make sure that all parkings that have time_end in the
future or not set will get the default domain as their domain, since
there might also be parkings that have not yet started, but they should
be updated still.  Also cast the domain area to MultiPolygon only if
it's not already one, since it might be, if the existing polygons were
separate.

In migration 0035, when creating the PaymentZone 3, use id as the lookup
key and other values as defaults so that it won't fail if id=3 already
exists.  Also make it use the zone with the greatest number as the base
where to copy the domain and area from.


suutari-ai:
@kerkkoheiskanen or @japauliina Can either of you review this, please?

PR #164, opened 2020-05-13
Fix active_permit_by_external_id series filtering

suutari-ai:
The active_permit_by_external_id endpoints in Enforcement API and
Operator API should operate on the latest active PermitSeries of the
user who is performing the request.  It used to operate on the latest
active PermitSeries regardless of the owner of the series, which made
that endpoint to create the permits to incorrect series some times.

Fix the endpoints by adding correct filtering to the base class
ActivePermitByExternalIdViewSet and add a test case as a regression test
for this bug.


suutari-ai:
Force-pushed as an attempt to get the codecov report work. Seemed to help.

PR #166, opened 2020-05-18
Admin: Improve Permit related views

suutari-ai:
Add more data visible to the permit related views in Django Admin and
add new filtering columns and a date hierarchy to permit lookup items.
Make domain code visible in string representation of PermitArea and
Permit objects.

Also add tests for these admin views and the added functionality.


suutari-ai:
Codecov has still some problems. Trying to close this and reopen from the same repo (i.e. not from fork). Maybe that will help.

suutari-ai:
Replaced with #167

PR #95, opened 2019-12-03
Implement endpoint for getting parking lot information

dvainio:
* Implement endpoint for querying information for parking
lots with the terminal number or coordinates.

* Implement function for fetching closest parking terminal
from coordinates within 100m.

* Implement tests.

* Add url.

PR #169, opened 2020-05-18
Couple fixes to Enforcement and Operator API documentation

suutari-ai:
Fix activepermit endpoint URLs.

Fix "end_date" field name to correct name "end_time".

Remove the parking_info_query endpoint documentation, since it's not
implemented yet.

Update the documentation versions to 1.2.0.rc3.  (Hopefully last rc
before actual release version of the docs.)


suutari-ai:
The parking_info_query endpoint is implemented in PR #95, but even that doesn't yet provide any data from the endpoint, since there's no data models to store it. When that work is finished, the endpoint can be re-added to documentation (by reverting commit 2770301).

PR #172, opened 2020-06-05
Archive parkings

japauliina:
The prevent the Parking table becoming too big, add a possibility to archive old parkings.


parkings/migrations/0038_auto_20200603_0748.py
```diff
@@ -0,0 +1,51 @@
+# Generated by Django 2.2.12 on 2020-06-03 07:48
```
suutari-ai, line 1: Please rename this migration so that it will have a better name than "auto..."

parkings/models/parking.py
```diff
@@ -42,17 +42,13 @@ def registration_number_like(self, registration_number):
...
-class Parking(TimestampedModelMixin, UUIDPrimaryKeyMixin):
+class AbstractParking(TimestampedModelMixin, UUIDPrimaryKeyMixin):
     VALID = 'valid'
     NOT_VALID = 'not_valid'
 
-    region = models.ForeignKey(
-        Region, null=True, blank=True, on_delete=models.SET_NULL,
-        related_name='parkings', verbose_name=_("region"),
-    )
+    region = models.ForeignKey(Region, null=True, blank=True, on_delete=models.SET_NULL, verbose_name=_("region"))
```
suutari-ai, line 49: I guess this is just for removing the related_name? It would be better to not touch the styling while removing this single argument.

I think the related_name  needs to be added to the Parking model then again.

suutari-ai: Oh, there was a default_related_name. Didn't know that such thing exists. Nice. :)

parkings/models/parking.py
```diff
@@ -152,7 +144,7 @@ def save(self, update_fields=None, *args, **kwargs):
             self.normalized_reg_num = (
                 self.normalize_reg_num(self.registration_number))
 
-        super(Parking, self).save(update_fields=update_fields, *args, **kwargs)
+        super(AbstractParking, self).save(update_fields=update_fields, *args, **kwargs)
```
suutari-ai, line 147: You can use the cleaner `super()` without arguments here, since this is a Python 3 code base.

parkings/utils.py
```diff
@@ -0,0 +1,18 @@
...
+def archive_parkings(time):
+    parkings_to_archive = Parking.objects.filter(time_end__lte=time)
+
+    for parking in parkings_to_archive:
+        with transaction.atomic():
+            fields = [field.name for field in AbstractParking._meta.fields]
+            values = dict([(field, getattr(parking, field)) for field in fields])
+            archived_parking = ArchivedParking(**values)
+            archived_parking.save()
+            parking.delete()
```
suutari-ai, line 16: I think this would be better as manager method in the ParkingsQuerySet.

I think the methods in the ParkingQuerySet could be something like this:
```python

# in the management command
    def handle(...)
        parkings_to_archive = Parking.objects.ends_before(before)
        if confirm:
             print(parkings_to_archive.count()) # etc.
        parkings_to_archive.archive()


class ParkingQuerySet(...):
    def ends_before(self, time):
        self.exclude(time_end=None).filter(time_end__lt=time)

    def archive(self):
        for parking in self:
            parking.archive()

class Parking(...):
    def archive(self):
        archived_parking = self.make_archived_parking()
        with transaction.atomic():
            archived_parking.save()
            self.delete()
            return archived_parking

    def make_archived_parking(self):
        fields = [x.name for x in self._meta.fields]
        values = {x: getattr(self, x) for x in fields}
        return ArchivedParking(**values)


class ArchivedParking(...):
    def archive(self):
        return self
```

suutari-ai:
Should modified_at and created_at be handled somehow specially? At least when I tried something similar locally I wasn't able to preserve those values.

IMHO it would be best to preserve the values of created_at and modified_at in the ArchivedParking.

suutari-ai:
The easiest way I can think of for preserving the created_at and modified_at timestamps of the archived parkings is something like this (if this just works -- didn't test):
```python
class ArchivedParking(AbstractParking):
    created_at = models.DateTimeField(verbose_name=_("time created"))
    modified_at = models.DateTimeField(verbose_name=_("time modified"))
    archived_at = models.DateTimeField(auto_now_add=True, verbose_name=_("time archived"))
```

I.e. removing the auto_now_add and auto_now from the fields. Also added a new field archived_at with auto_now_add=True so that the time of archiving is available.

parkings/tests/test_archive_parkings.py
```diff
@@ -0,0 +1,43 @@
...
+
+    assert Parking.objects.all().count() == 0
+    assert ArchivedParking.objects.all().count() == 1
+
+    archived_parking = ArchivedParking.objects.first()
+
+    assert archived_parking.id == parking.id
+    assert archived_parking.registration_number == parking.registration_number
+    assert archived_parking.time_end == parking.time_end
+    assert archived_parking.location == parking.location
```
suutari-ai, line 43: add modified_at and created_at to this list

parkings/models/parking.py
```diff
@@ -79,18 +73,16 @@ class Parking(TimestampedModelMixin, UUIDPrimaryKeyMixin):
...
+        PaymentZone, on_delete=models.PROTECT,
         verbose_name=_("PaymentZone"), null=True, blank=True)
     is_disc_parking = models.BooleanField(verbose_name=_("disc parking"), default=False)
 
     objects = ParkingQuerySet.as_manager()
 
     class Meta:
-        verbose_name = _("parking")
-        verbose_name_plural = _("parkings")
+        abstract = True
```
suutari-ai, line 85: Maybe the methods after this could be on the Parking model rather than in the AbstractParking? Just thinking that ArchivedParking should probably not have the customization of the save method and they could have a different `__str__`.  Or the `__str__` could be in the base class and then just call super in the `ArchivedParking.__str__` and add string `archived` to it, or similar.

PR #182, opened 2020-08-06
Fix pagination of valid_permit_item endpoint in Enforcement API

suutari-ai:
The pagination of valid_permit_item endpoint in the Enforcement API
didn't work correctly.  When querying all the pages, the results didn't
always contain all items, but some items "between" the pages were
excluded.  Fix this issue by switching to cursor pagination ordered by
the id field (which is always increasing).

Also improve a few related things:

  * Add id and permit_id to the result so that it's easier to verify
    that all is there.

  * Relax the required parameters, because it is no longer needed to
    have the time parameter for pagination.

  * Minor improvement to code readability in get_queryset.

### DIFF
--- a/docs/api/enforcement.yaml
+++ b/docs/api/enforcement.yaml
@@ -503,8 +503,15 @@ paths:
                     - << : *validPermitItemExample
                       area: "B"
                       registration_number: "XYZ-456"
-                properties:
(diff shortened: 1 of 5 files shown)


suutari-ai:
@japauliina I'm going to have to self merge this, but just FYI.

PR #194, opened 2021-03-16
Anonymize registration numbers

amanpdyadav:
Anonymization of registration number for the following models:
    * Parking
    * Permit
    * ParkingCheck
    
Once the grace period of 24 hours is over the registration numbers
are anonymized by assigning it an empty string.


parkings/anonymization/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    for parking_check in parking_checks:
+        with transaction.atomic():
+            parking_check.registration_number = ""
+            parking_check.save(update_fields=["registration_number"])
+
+
+def anonymize_permit_registration_numbers():
+    parking_lookups = (
```
tuomas777, line 36: Nitpick ofc, but I guess this variable should be named `permit_lookups`.

parkings/anonymization/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+            subjects = permit.subjects
+            for sub in subjects:
+                sub["registration_number"] = ""
+            permit.save(update_fields=["subjects"])
+
+
+def get_past_time():
+    time = timezone.now()
+    grace_duration = getattr(settings, "PARKKIHUBI_ANONYMIZE_REG_NUM_BEFORE", None)
+    past_time = time - grace_duration
```
tuomas777, line 54: IMO it is better to remove the fallback value and let this fail on the `getattr()` rather than on the next line with some not so clear exception.

parkings/tests/test_parkings_anonymization.py
```diff
@@ -0,0 +1,93 @@
...
+        subjects = permit.subjects
+        for sub in subjects:
+            sub["end_time"] = str(end_time)
+            permit.save(update_fields=["subjects"])
+
+    call_command(clean_reg_nums.Command())
+
+    assert ParkingCheck.objects.filter(registration_number="").count() == result
+    assert Parking.objects.filter(registration_number="").count() == result
+    assert PermitLookupItem.objects.exclude(registration_number="").count() == permit_result
```
tuomas777, line 93: All of these tests seem to check outcomes that either all of the objects are changed, or that none of them is changed. I think optimally we would like to verify that the anonymize functions can also change just some of the objects, and leave others unchanged on one run.

tuomas777 (commented):
Added a couple of mostly opionated and/or nit-pickish comments 🙂 In general looks very good to me 👍 There seems to be one failing test though(?)

parkings/anonymization/__init__.py
karisal-anders: Will take into account.

parkings/anonymization/__init__.py
karisal-anders: Will take into account.

parkings/tests/test_parkings_anonymization.py
karisal-anders: Ok, a test case where some objects are anonymized and some are not is not present currently. Due to time constraints I probably won't add such a case if it is not strictly necessary.

karisal-anders:
@tuomas777 Replied to all of your comments above. I will check that the tests pass locally, I think at the moment the GitHub's automatic testing is not working in this project.

parkings/anonymization/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+def anonymize_parking_registration_numbers():
+    parkings = (
+        Parking.objects.filter(time_end__lt=get_past_time())
+        .exclude(registration_number="")
+        .order_by("time_end")
+    )
+
+    for parking in parkings:
+        with transaction.atomic():
```
tuomas777, line 16: Spotted one thing I missed on the first round; these atomic transactions don't seem to make any difference as there is only one save operation inside them. I wonder if the idea was to run all of the changes in a transaction (which would probably make sense) and this is just in the wrong place.

karisal-anders: I don't know what the idea was because I am not the author of this code but I concur that having a transaction.atomic() around a single object save is not necessary here. I can move the transaction clauses outside each for loop, e.g.:
```python
for parking in parkings:
        with transaction.atomic():
            parking.registration_number = ""
            parking.normalized_reg_num = ""
            parking.save(update_fields=["registration_number", "normalized_reg_num"])
```
→
```python
with transaction.atomic():
    for parking in parkings:
            parking.registration_number = ""
            parking.normalized_reg_num = ""
            parking.save(update_fields=["registration_number", "normalized_reg_num"])
```
so then either all of them go through or none of them go through.
Would that be an acceptable solution?

karisal-anders: Having now talked with both the author of this pull request and @tuomas777 we came to the consensus that removing the ```with transaction.atomic():``` clauses is here is ok.

parkings/anonymization/__init__.py
karisal-anders: Taken into account in PR #223 by removing the variable name completely (See https://github.com/City-of-Helsinki/parkkihubi/pull/223/commits/31be2a131ce5a7b815073d6350c626ec0ed54b52#diff-125eddeefa8c7da8082d77756549474a074b240ea65e28f2b6c190f568d64dafL32-R16).

parkings/anonymization/__init__.py
karisal-anders: Taken into account in PR #223 (See [54aa1a0e2c5a058e7db0666c1b6e02529610bd70](https://github.com/City-of-Helsinki/parkkihubi/pull/223/commits/54aa1a0e2c5a058e7db0666c1b6e02529610bd70)).

parkings/anonymization/__init__.py
karisal-anders: Taken into account in PR #223 (See https://github.com/City-of-Helsinki/parkkihubi/pull/223/commits/6de374eb3df75222cddb6c16db7d9c80968f792d).

karisal-anders:
> > Added a couple of mostly opionated and/or nit-pickish comments 🙂 In general looks very good to me 👍 There seems to be one failing test though(?)
> 
> @tuomas777 Replied to all of your comments above. I will check that the tests pass locally, I think at the moment the GitHub's automatic testing is not working in this project.

The changes are ready to be inspected in PR #223, please look there—thank you!

suutari-ai:
Merged the #223.

PR #227, opened 2022-02-04
Implement parkings data export capability

HotStew:
### Export button to toggle options rendering
image
______________
### Nothing selected, start & end time automatically set. 
image
______________
### Filters selected
image
______________
### CSV file
image
______________

### Task requirements:
Export parkings data to CSV file.
* Fields included in CSV must be: Location coordinates, terminal number, start time, end time, created at -time, modified at -time, operator, payment zone.
* User may filter data to be exported with: operator(multi select), payment zone(multi slect), time range (start & end), parking checked.
* For now, filter by activity in time range. So if parking was active anytime **within** the time range.

### DIFF
--- a/parkings/api/monitoring/export_parkings.py
+++ b/parkings/api/monitoring/export_parkings.py
@@ -1,0 +1,112 @@
+import csv
+
+from rest_framework import renderers, serializers, viewsets
+from rest_framework.decorators import action
+from rest_framework.response import Response
+from six import StringIO
+
+from ...models import Parking, ParkingCheck
+from .permissions import IsMonitor
+
+
+class CSVRenderer(renderers.BaseRenderer):
+    media_type = "text/csv"
+    format = "csv"
+    render_style = 'binary'
+
+    def render(self, filters, media_type=None, renderer_context=None):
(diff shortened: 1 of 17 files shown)


suutari-ai:
Oops, closed this by accident while closing some some cruft PRs in bulk.

suutari-ai:
You shouldn't use f-strings, since they are not supported with Python 3.5.

Please fix the style issues too.

parkings/api/monitoring/export_parkings.py
```diff
@@ -0,0 +1,112 @@
...
+
+        if parking_check:
+            parking_check_parking_ids = (
+                ParkingCheck.objects
+                .filter(found_parking__id__in=parkings)
+                .order_by("found_parking")
+                .values_list("found_parking")
+                .distinct()
+            )
+            parkings = parkings.filter(id__in=parking_check_parking_ids)
```
suutari-ai, line 49: The filtering code here doesn't seem to belong to the CSV renderer, but rather to the ViewSet, or even better they should be implemented as FilterSets. See e.g. parkings/api/enforcement/valid_parking.py for an example of FilterSet.

PR #237, opened 2022-05-17
Add github actions to run the tests

amanpdyadav:
Refs None

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,0 +1,28 @@
+name: CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


suutari-ai:
1. You have to create a branch to the main repo for this. Otherwise GitHub won't run the jobs.
2. Also remove the Travis configuration, and
3. add codecov upload task. There is an example at [1], but the version of it should be updated from v1 to v3 (see [2]).

[1] https://github.com/andersinno/django-analog/blob/master/.github/workflows/test.yml#L32-L33
[2] https://github.com/marketplace/actions/codecov

suutari-ai:
Create another PR from a branch in this repo to make the actions run. See my previous comment for other details.

PR #238, opened 2022-05-17
Replace Travis with GitHub Actions

amanpdyadav:
* Add github actions to run the tests
* Remove travis confirguration


.github/workflows/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
+  test:
+
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-latest]
```
suutari-ai, line 14: Maybe we don't need macos or windows for now? Or we can have them too, if the tests don't take very long and they pass. Let's see how much they prolong the CI times...

.github/workflows/main.yaml
```diff
@@ -0,0 +1,31 @@
...
+        python: ["3.8", "3.9", "3.10"]
```
suutari-ai, line 15: I think we need to have at least Python 3.5 too. OTOH 3.8 might not be needed.

.travis.yml
```diff
@@ -1,45 +0,0 @@
...
-      python: "3.5"
-      addons: null
-      before_script: null
-    - name: Coding Style
-      env: TOXENV=style
-      python: "3.5"
-      addons: null
-      before_script: null
-    - name: Sanitizer Configuration
-      env: TOXENV=sanitizer
```
suutari-ai, line 25: These three "Requirements", "Coding Style", and "Sanitizer Configuration" should be preserved in the GitHub Actions.

Something like this should work for them:
https://github.com/andersinno/django-analog/blob/eb5f481cbf399c359fb52174a184ace0bab71da7/.github/workflows/test.yml#L15-L17

suutari-ai:
It's not actually running anything for all envs. See:
image

I think you're missing the tox-gh-actions configuration in tox.ini. See https://github.com/andersinno/django-analog/blob/eb5f481cbf399c359fb52174a184ace0bab71da7/tox.ini#L21-L26

.github/workflows/main.yaml
amanpdyadav: `ubuntu-latest` and `macos-latest` seem to run quite fast but I don't know if windows are even running? So should we remove it?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  test:
+
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, macos-latest, windows-latest]
+        python: ["3.5", "3.9", "3.10"]
+        toxenv: [ "" ]
+        include:
+          - { platform: ubuntu-latest, python: "3.5", toxenv: "style" }
```
suutari-ai, line 18: Add also the requirements and sanitizer

suutari-ai (approved):
Thanks Aman for pushing this!

Had to do quite many changes to the workflow file, but got it working finally.

The tests finally work again. Yay!

Let's merge this.

PR #240, opened 2022-05-20
Permit properties

amanpdyadav:
* Remove permit_type from the documentation.
* Add permit_type to the valid permit list
* Add example of permit_type


suutari-ai:
Shouldn't this need implementation too?

amanpdyadav:
> Shouldn't this need implementation too?

Some how my editor missed one commit to save. Should be there now.

suutari-ai (approved):
Looks good now. Thanks!

PR #241, opened 2022-05-31
Remove package-lock.json as yarn is used

amanpdyadav:
Yarn is used as a package manager and therefore
package-lock.json file is not necessary.

Refs None


suutari-ai (approved):
Thanks!

PR #269, opened 2023-11-15
Bump docker postgis

samimedallal:
Production was upgraded to PostgreSQL 15.4 so docker postgis needs to be upgraded as well to the closest version.


suutari-ai (approved):
Thanks Sami! :+1:

PR #276, opened 2023-11-24
Multiple allowed users for PermitArea (HPH-36)

suutari-ai:
Add support for allowing sharing of the PermitAreas between operators and the enforcer.

Previously it was possible to use each PermitArea for just a single enforcer or operator.
This was especially problematic, because the PermitArea codes were unique within
enforcement domains so it was not even possible to create duplicate areas with the
same identifiers.


mhieta (approved):
LGTM!

mhieta (approved):
Good!

